PR #4, opened 2012-11-27
Confusing List::Util::first() documentation

mavit:
I believe that the explanation of `first()` in terms of `reduce()` is meant to read:

     $foo = reduce { wanted($a) ? $a : wanted($b) ? $b : undef } undef, @list


leonerd:
No; we don't apply the wanted() function repeatedly to $a, only $b.
In any case this was rewritten some time ago; closing.

PR #11, opened 2013-11-27
Update synopsis to list all exportable functions

karenetheridge:


leonerd:
Rewritten for current list as 0b986b9

PR #17, opened 2014-08-26
skip sum IV precision test on older perls (<5.8)

rurban:
with 5.6 sum will still use NV, and looses 53-64bit precision.

this patch is on top of your sub-util and my reduce-smg


leonerd:
This diif no longer applies cleanly; there seems to be lots of other changes that have crept in.

Can you recreate it?

leonerd:
Superceded by #29

PR #20, opened 2014-12-11
Improve SYNOPSIS

basiliscos:
Hello!

This is one of my most commonly-used modules which I use more then year. What was my surprise, when I found that there are wonderful pair-functions. 

I think they (and all other functions) should be mentioned in SYNOPSYS.

Thanks for good module!


leonerd:
Solved by 0b986b9

PR #23, opened 2015-05-13
add head() and tail() subs

preaction:
`head( $n, @list )` grabs the first `$n` elements from `@list`, or if `$n` is
negative, all but the last `$n` elements.

`tail( $n, @list )` grabs the last `$n` elements from `@list`, or if `$n` is
negative, all but the first `$n elements.

These are useful if you need to slice a list and do not want to assign
it to a temporary array. For example, as part of a map/grep/sort/etc...
chain.

---

With this setup, a slice() sub would be pretty trivial. If this stuff isn't appropriate for List::Util, I could make them into their own dist.

### DIFF
--- a/ListUtil.xs
+++ b/ListUtil.xs
@@ -559,6 +559,56 @@ PPCODE:
 }
 
 void
+head(size,...)
+PROTOTYPE: $@
+ALIAS:
+    head = 0
+    tail = 1
+PPCODE:
+{
+    int size = 0;
+    int start = 0;
+    int end = 0;
+    int i = 0;
+
+    size = SvIV( ST(0) );
+
+    if ( ix == 0 ) {
+        start = 1;
+        end = start + size;
+        if ( size < 0 ) {
+            end += items - 1;
+        }
+        if ( end > items ) {
+            end = items;
+        }
+    }
+    else {
+        end = items;
+        if ( size < 0 ) {
+            start = -size + 1;
+        }
+        else {
+            start = end - size;
+        }
+        if ( start < 1 ) {
+            start = 1;
+        }
+    }
+
+    if ( end < start ) {
+        XSRETURN(0);
+    }
+    else {
+        EXTEND( SP, end - start );
+        for ( i = start; i <= end; i++ ) {
+            PUSHs( sv_2mortal( newSVsv( ST(i) ) ) );
+        }
+        XSRETURN( end - start );
+    }
+}
+
+void
 pairs(...)
 PROTOTYPE: @
 PPCODE:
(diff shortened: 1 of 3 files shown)


preaction:
This was released to CPAN as List::Slice, which includes some bugfixes to prevent stack corruption and eventual segfaults. I'll update this pull request with the new code tonight, just to keep everything up-to-date for when we figure out whether it should go in here or not.

leonerd:
Those seem a nice idea. I'm reluctant to pull things into the real core L:U module for now though for such a new idea, even how simple and obvious it is.
Lets see how List::Slice fares on CPAN and consider it for a merge into L:U proper in a little while.

leonerd:
How are these doing currently? I see they're on CPAN as their own dist - does that look like it's gaining traction? Any interface issues we should consider?

preaction:
The blog post seemed to go over well, there's a couple upvotes on metacpan, and a couple people on IRC have expressed their interest, but I think it's likely trivial enough that adoption would take a while to see (the weight of a dependency, XS even, is a lot larger than just making a temporary array and slicing that).

At the very least, I need to update this PR with bugfixes: My logic is wrong, `XSRETURN( end - start )` can go negative and destroy the save stack. I'll do that this week.

preaction:
Updated and rebased. This is what's in the List-Slice dist.

pilcrow:
The downside to the current XS implementation is that, in a `for` loop, the slice is copied to an anonymous array, meaning that `head($lots, ...)` will copy `$lots` of elements (if they exist).  By contrast, a conventional slice in a for loop fetches elements one at a time (e.g., `for (@array[0..$n])`).

If you restricted this function to _arrays_ rather than lists, you could get the same effect.  Something like:

    sub head($\@) {
      my ($n, $ary) = @_;
      $n = @$ary + $n if $n < 0;
      @$ary[0..min($#$ary,$n-1)];  # won't be copied in a for loop
    }

You can confirm the behavior with some `tie`d array that barks on every FETCH.

Grinnz:
That the functions can operate on lists and not just arrays is essentially the purpose for their existence.

preaction:
But if the behavior was restricted to arrays, you can already use an array slice. head/tail have more value when you don't want to make an array, and you don't want to circumfix a slice syntax around a set of map/grep/sort.

```perl
my @top_five = sort head 5, grep { $_->active } @users;
```

preaction:
I should also mention I tried just slicing the stack directly, but Perl did not like that. I thought I could get away with it for `head`, even if I couldn't for `tail`, but it did not work (I don't recall precisely why, but it was certainly about the ref count). I'm pretty sure this was before I fixed the stack-destroying, negative-`XSRETURN` bug, so it might still be possible

pilcrow:
@Grinnz , yes, of course.  That was more of a lament.

leonerd:
Done :)

PR #27, opened 2015-06-23
test product(0, 0) - RT#105415

chorny:


leonerd:
Not merging because it conflicts, but reïmplemented in 4372433

PR #28, opened 2015-07-04
Copy uniq() implementation and tests from List::MoreUtils

autarch:


autarch:
It'd be really nice if this repo was hooked up to travis.

autarch:
Comments? Thoughts? Concerns?

leonerd:
Did we decide on having an `nuniq` too? Use the `ALIAS` feature to decide if using string or numerical comparison

Grinnz:
Does this implementation consider '1' and 1 to be the same? The "undef" handling is different from List::UtilsBy::uniq_by and may be surprising. Either way, the description should clearly define how elements are considered for uniqueness, as List::UtilsBy::uniq_by does.

autarch:
I didn't write the original code, but based on looking at it, it just stores each value in a hash to uniquify them, so it'd treat `1` and `"1"` the same way. This is generally consistent with Perl itself. Treating `undef` as its own value seems about as sane as any other option. As far as acting like List::UtilsBy, I don't know why that implementation is the reference point. FWIW, List::MoreUtils has been around for ages and I think a lot of people have used it in the past.

Grinnz:
I didn't mean to suggest L::UB is "the" reference point, just "a" reference point. If it tests uniqueness as hash keys but with special handling for undef, that's fine with me, I just want to make sure that gets documented.

karenetheridge:
:+1: to getting uniq into List::Util!  I've been copying the pure-perl implementation from LMU all over the place and it would be great to be able to remove that.

leonerd:
Ah. Oops. I had forgotten this was a PR that existed. I've already recreated uniq() and also added a uniqnum(), now sitting on the `uniq` branch, which is hoping to become 1.44.

ribasushi:
@leonerd The implementation of `uniq()` as present in the [latest devel 1.44_001](https://metacpan.org/diff/file?source=PEVANS/Scalar-List-Utils-1.43&target=PEVANS/Scalar-List-Utils-1.44_001) fails a test in this PR:
```
#   Failed test 'none of the numbers in the array were stringified by uniq'
#   at t/uniq.t line 44.

#   Failed test 'uniq uniquifies mixed numbers and strings correctly in a tied array'
#   at t/uniq.t line 58.
#     Structures begin differing at:
#          $got->[1] = Does not exist
#     $expected->[1] = '2'
```

/cc @shadowcat-mst @kentfredric @ap @haarg

leonerd:
Hmmm.. Hard to say what's the best thing to do with those failures.

The first of them (about not stringifying) tests the very fragile property of the `SvPOK` flag on elements of the input list. That flag is an internal-representation detail and I don't feel is the sort of thing worth worrying about. For example, simply interpolating elements into an array to print it would side-effect that flag on (think `print STDERR "Debug: values are <@values>\n";`)

The tied array test looks potentially interesting though. Perhaps some missing magic calls in my implementation. I'll take a look.

autarch:
Should this PR be closed and the discussion moved elsewhere?

leonerd:
Yes - that might be best in fact. https://rt.cpan.org/Ticket/Display.html?id=113294

PR #29, opened 2015-08-10
rebasing of rurban PR #17 (skip sum IV precision test on <5.8)

Grinnz:
Skip last test in sum.t on 5.6, from #17 and https://rt.cpan.org/Public/Bug/Display.html?id=98323


leonerd:
Eh. I dislike simply skipping tests that appear to fail in some setups, without at least *some* understanding of why.

Grinnz:
This is no longer needed after #55

PR #30, opened 2015-08-21
Remove C99 comment, it was failing on C89 compilers. Also, the array …

andygrundman:
…ref checking is already being done in the previous line of code.


leonerd:
Superceded by #31

PR #32, opened 2015-10-28
Remove unused variables in pairmap()

ilmari:
When `List::Util` is built as part of core, `-Wunused-vars` gets set, and
gcc complains about these variables.

They were added by the commit 8630ea87 on 2013-08-01, but never used.


leonerd:
LGTM

PR #33, opened 2015-11-19
Remove stale documentation, add commas, article, etc.

benkasminbullock:
Regarding this:

https://rt.cpan.org/Public/Bug/Display.html?id=108558

I suggest removing the section.

Also added commas and missing article.


adamherzog:
:+1:

leonerd:
Heh. Still not ideal - $! isn't a tied scalar; it's magical. I'll fix that one up

PR #34, opened 2015-12-03
Add a LICENSE file.

shlomif:
Hi all! Please merge this commit with a "LICENSE" file.

It is for CPANTS Kwalitee, GitHub etiquette, good measure and for Great
Justice!

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,379 @@
+This software is Copyright (c) 1997-2007 by Graham Barr.
+
+This is free software; you can redistribute it and/or modify it under
+the same terms as the Perl 5 programming language system itself.
+
+Terms of the Perl programming language system itself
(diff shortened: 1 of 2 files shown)


adamherzog:
:+1:

leonerd:
I dislike this.

Does EU::MM not have an equivalent to Module::Build's "just make me a LICENCE file" option? I use that and don't have to worry about this issue in my other >100 dists.

shlomif:
@leonerd: I'm not aware of such an option in EU::MM but note that I try to avoid raw EU::MM as much as possible and prefer Module::Build and especially Dist-Zilla. I don't see anything for generating a LICENSE file here - https://metacpan.org/pod/ExtUtils::MakeMaker .

PR #35, opened 2015-12-03
Manifest.skip ignore vim backup files

shlomif:


adamherzog:
:+1:

leonerd:
This is an unclean diff as it also applies the LICENCE file.
However I've recreated the intention of it in 9cec0c9

PR #36, opened 2016-01-27
Addresses the two outstanding Kwalitee issues

neilb:
Hi Paul,

This adds the min perl version to the dist's metadata, via `MIN_PERL_VERSION` in `Makefile.PL`.

And I added `use warnings` to the modules that didn't already have it.

Cheers,
Neil


adamherzog:
:+1:

leonerd:
Seems good.

PR #37, opened 2016-01-31
Add Travis CI config.

adamherzog:
This adds a pretty basic config file for Travis CI, for automatic testing against stable Perls from 5.8 through 5.22. (Travis does not support Perl 5.6.)

I think this would be helpful in providing some immediate feedback and help both the maintainer and future contributors.

You'll need to setup an account at Travis-CI to use this, of course, but it's free and quick to setup.


leonerd:
I'm really not sure I understand this one. Can you explain it a little more?

adamherzog:
Travis CI is a continuous integration service, that provides free CI services to open repositories like this one. It allows for automatic building and testing of the module after each commit, and for each pull request. The module is built and tested against a number of different versions of Perl - currently configured for 5.8 through 5.22. The entire build and test process can be customized via the .travis.yml configuration file, though this module is pretty straightforward so there isn't much customization here. Details on the build options can be found here: https://docs.travis-ci.com/user/languages/perl

It integrates with Github to display a small icon next to each branch/PR, indicating if the build passed/failed. The main value here, I think, is that it allows both the maintainer and any contributors to quickly and easily see whether a commit/PR passes or fails tests.

For an example, see the repo for Net-NetMask: https://github.com/muir/Net-Netmask

You'll see a small green checkmark or a red x next to each branch/PR. See https://github.com/muir/Net-Netmask/branches and https://github.com/muir/Net-Netmask/pulls

By clicking on those icons, you'll be taken to Travis-CI where you can view the build details, or view all of the builds. When a build fails, you can view the actual build log to see what went wrong. For example: https://travis-ci.org/muir/Net-Netmask/builds/61245757

The primary deficiency (IMO) of Travis is that it offers a limited selection of platforms. It will test against 64 bit Linux VMs, only, I believe. So we aren't getting coverage on other architectures or platforms, but it will still help ensure that the build always succeeds and tests pass against a variety of Perls, before pulling a PR, for example.

leonerd:
I still don't follow. Do I have to run a travis instance somewhere? Or is this some magical thing that runs elsewhere and just works.. somehow..?

adamherzog:
It's a service that lives somewhere else and just works. But you do need to sign in to the Travis-CI website and enable this repo.

Getting it set up should look something like this: go to https://travis-ci.org/Scalar-List-Utils/Scalar-List-Utils and login using your GitHub credentials (it uses oauth, I believe - you don't need to register yet another account.)

It'll ask for permission to view your repositories. Assuming you say okay, Travis will sync a list of your repos over. Then you'll be able to enable Travis CI on a per-repo basis. Enable the services on this repo.

Once that's done, you're done. The next time a commit is made or a new branch is created, it should automatically clone the repo, build the dist, and run the tests, against a bunch of different versions of perl. The results of the build process will appear next to the commit on GitHub, as well as in PRs, I believe.

(If you're interested in having this set up, but don't want to deal with actually setting it up, I'd be happy to do it for you. You'd need to add me to the Scalar-List-Utils organization, at least temporarily.)

leonerd:
LGTM

pali:
Looks like Travis tests are not working anymore... @adherzog any idea?

adherzog:
@pali It's probably because the organization name changed - the repo used to be Scalar-List-Utils/Scalar-List-Utils, but now it's Dual-Life/Scalar-List-Utils.

I think somebody from the Dual-Life organization will need to add the repo to Travis-CI.

PR #15, opened 2014-08-03
Topic/fix metadata

karenetheridge:
A few small fixups to the dist, after I noticed its kwalitee report...

PR #38, opened 2016-01-31
Prep for 1.43 release

adamherzog:
I've bumped the version number and populated the Changes file, to help prepare the next release.

I followed the basic format already used in the Changes file, but also indicated the PR # if the change came from a contributor on GitHub and not through RT. There have also been some changes to the dist MANIFEST and Makefile.PL, but I've omitted those from the list of changes.

There are also 4 open PRs right now that I've recently +1ed and could be quickly pulled before this release, if you agree:

 * #33 includes documentation changes only.
 * #34 and #35 are dist related changes only.
 * #36 is also dist related, but also includes adding 'use warnings' to each module. That had previously been suggested in #15 as well. As karenetheridge noted, It should be safe to do given the minimum perl of 5.006.

If you do want to pull any of those those (or any other PRs) before the next release, I'd be happy to update the Changes file accordingly.

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,3 +1,16 @@
+1.43 -- 2016/XX/XX XX:XX:XX
+	[CHANGES]
+	 * Documentation improvements in List::Util
+
+	[BUGFIXES]
+	 * Fix build on C89 compilers (e.g. VC++) (PR24)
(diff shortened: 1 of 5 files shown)


leonerd:
The 1.43 release was already made. We're currently testing 1.43_devel releases to add uniq() and uniqnum().

PR #39, opened 2016-03-10
Don't reference newsp on >= 5.23.8 (RT#112312)

ilmari:
dMULTICALL no longer declares this variable


karenetheridge:
:+1:

jhi:
It would be great if SLU 1.44 could be minted with this change, so that bleadperl could be updated with the latest SLU fixes.

PR #41, opened 2016-04-24
Fix inline variable declaration for non-C99 compilers

andygrundman:
MSVC is a real pain... here's a patch to fix another inline variable declaration.


bulk88:
Without this patch, SLU can not be built with Win32 Visual C compiler.

PR #42, opened 2016-05-16
Fix building with C++ compilers

arc:
This change is (about to be) in blead as customized 1.45_01.


arc:
And a second commit just added, to fix building under strict-C89 compilers.

PR #43, opened 2016-05-22
Add pick() function to pick a single random item from the input list

preaction:
This PR adds a `pick()` function that will pick a single random item from the input list. This is like the [Perl 6 pick routine](https://doc.perl6.org/routine/pick), except the count is always 1 (I can add a count argument if desired).

This is faster than using `shuffle()` and picking the first one, and is more convenient for picking a number from a range:

```
# Pick a number from 10-99
my $num = int( rand 90 + 10 );

use List::Util qw( pick );
my $num = pick 10..99;
```

### DIFF
--- a/ListUtil.xs
+++ b/ListUtil.xs
@@ -1041,19 +1041,42 @@ CODE:
 }
 
 void
-pick(...)
-PROTOTYPE: @
+pick(num, ...)
+    int num;
+PROTOTYPE: $@
 CODE:
 {
     my_init_rand();
 
-    if ( items <= 0 ) {
-        XSRETURN_UNDEF;
+    if ( num <= 0 || items <= 1 ) {
+        XSRETURN_EMPTY;
     }
 
-    int index = (int)(Drand01() * (double)(items));
-    ST(0) = ST(index);
-    XSRETURN(1);
+    if ( num > items - 1 ) {
+        num = items - 1;
+    }
+
+    int found[num];
+    int i = 0;
+    while ( i < num ) {
+        bool redo = false;
+        int index = (int)(Drand01() * (double)(items-1));
+        for ( int x = 0; x < i; x++ ) {
+            if ( found[x] == index ) {
+                redo = true;
+                break;
+            }
+        }
+        if ( redo == true ) {
+            continue;
+        }
+
+        ST(i) = ST(index+1);
+        found[i] = index;
+        i++;
+    }
+
+    XSRETURN(i);
 }
 
 void
--- a/ListUtil.xs
+++ b/ListUtil.xs
@@ -114,6 +114,32 @@ static enum slu_accum accum_type(SV *sv) {
 /* Magic for set_subname */
 static MGVTBL subname_vtbl;
 
+/* Initialize Perl's rand() function */
+void my_init_rand() {
+#if (PERL_VERSION < 9)
+    struct op dmy_op;
+    struct op *old_op = PL_op;
+
+    /* We call pp_rand here so that Drand01 get initialized if rand()
+       or srand() has not already been called
+    */
+    memzero((char*)(&dmy_op), sizeof(struct op));
+    /* we let pp_rand() borrow the TARG allocated for this XS sub */
+    dmy_op.op_targ = PL_op->op_targ;
+    PL_op = &dmy_op;
+    (void)*(PL_ppaddr[OP_RAND])(aTHX);
+    PL_op = old_op;
+#else
+    /* Initialize Drand01 if rand() or srand() has
+       not already been called
+    */
+    if(!PL_srand_called) {
+        (void)seedDrand01((Rand_seed_t)Perl_seed(aTHX));
+        PL_srand_called = TRUE;
+    }
+#endif
+}
+
 MODULE=List::Util       PACKAGE=List::Util
 
 void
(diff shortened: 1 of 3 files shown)


leonerd:
This has a small conflict that will need fixing

preaction:
Fixed

grr:
Ability to pecify count would we much more useful. Grepping through my own codebase, I find I need pick > 1 about 50% of the time. And googling for "perl shuffle random pick" shows many results where the user wants to pick with count > 1.

preaction:
I've added the `$count` argument to the `pick()` function, since it does make it more generally useful. If we wanted, we could do a `picka`/`pick_a`/`pick1` or something to make it 1 thing.

grr:
Saw this similar request on rt: https://rt.cpan.org/Public/Bug/Display.html?id=105275
There's an interesting comment about improving efficiency by not using a complete shuffle.

grr:
I think this algorithm can be made more efficient. As is, the worst case could spend a lot of time in the inner loop checking if the current element was previously picked. Also the extra storage for the found array wouldn't be needed. The comment on this ticket mentions the better algorithm: https://rt.cpan.org/Public/Bug/Display.html?id=105275 . Perform a truncated shuffle, requiring only $count swaps. Note the guts of the current implementation could be repurposed to add a role() function as mentioned in this ticket: https://rt.cpan.org/Public/Bug/Display.html?id=120798

y:
Truncated shuffle version: https://github.com/y/Scalar-List-Utils/commit/2303638d665cf53a402c04246bdf1a46bf0b6625

grr:
i benchmarked the versions from both @preaction and @y and it looks like there is an exponential performance drop in preaction's version:

    benchmarking 1 picks / 10 elements:
                Rate         y preaction
    y         5415665/s        --       -1%
    preaction 5450518/s        1%        --

    benchmarking 10 picks / 10 elements:
                Rate preaction         y
    preaction 1146879/s        --      -42%
    y         1960478/s       71%        --

    benchmarking 10 picks / 100 elements:
                Rate preaction         y
    preaction 1390157/s        --       -3%
    y         1434347/s        3%        --

    benchmarking 100 picks / 100 elements:
                Rate preaction         y
    preaction  48188/s        --      -83%
    y         283881/s      489%        --

    benchmarking 100 picks / 1000 elements:
                Rate preaction         y
    preaction 113777/s        --      -41%
    y         192752/s       69%        --

    benchmarking 1000 picks / 1000 elements:
                Rate preaction         y
    preaction   621/s        --      -98%
    y         29537/s     4654%        --

    benchmarking 1000 picks / 10000 elements:
                Rate preaction         y
    preaction  3251/s        --      -82%
    y         18101/s      457%        --

    benchmarking 10000 picks / 10000 elements:
                Rate preaction         y
    preaction 4.63/s        --     -100%
    y         2791/s    60178%        --

    benchmarking 10000 picks / 100000 elements:
                Rate preaction         y
    preaction 37.3/s        --      -97%
    y         1174/s     3051%        --

    benchmarking 100000 picks / 100000 elements:
                s/iter preaction         y
    preaction     25.6        --     -100%
    y         6.64e-03   385920%        --

code: 

    #!/usr/bin/env perl
    use strict;
    use warnings;

    use List::Util qw(pick_y pick_preaction);

    use Benchmark qw(:all);

    for my $size (10, 100, 1_000, 10_000, 100_000) {
        my $count = $size / 10;
        print "\nbenchmarking $count picks / $size elements:\n";
        my @list = (0 .. $size);

        cmpthese -1, {
            y         => sub { my @a = pick_y         $count, @list },
            preaction => sub { my @a = pick_preaction $count, @list },
        };

        print "\nbenchmarking $size picks / $size elements:\n";
        cmpthese -1, {
            y         => sub { my @a = pick_y         $size, @list },
            preaction => sub { my @a = pick_preaction $size, @list },
        };
    }

dur-randir:
@grr, it's just quadratic compared to linear. If you add a no-op function in comparison, that'd just pass N elements and do nothing, you'll see that version from @y is indeed linear from the number of elements' picked, so I'm up for it.

But what I like in the original proposal is having a simple form when I just need one element - maybe let pick() always return one, and leave only sample() taking the additional num argument? What do you think?

grr:
> maybe let pick() always return one, and leave only sample() taking the additional num argument

That would probably be confusing. @preaction earlier suggested using pick1 for that case which is unambiguous.

BTW, I just noticed that List::MoreUtils added the samples() function last July. The timing and code similarity suggest @rehsack was aware of this ticket.

rehsack:
@grr as documented I implemented samples based on the wish in RT#106401. I did not research where the wish comes from. It sounds reasonable so I did it.

dur-randir:
@grr, just document what each of them does? Like

```
my $scalar = pick @list; # chooses one element from @list at random
my @res = sample 10, @list; # chooses ten elements from @list at random
```

I really like the possibility to omit '1, ' in the simple case - feels so much perlish.

Grinnz:
Not being aware of this proposal I opened https://rt.cpan.org/Ticket/Display.html?id=131535, and there is one question I raised there that's relevant: should the function that chooses multiple items choose with replacement? I suggested that it should not, as choosing multiple items with replacement can be reimplemented by calling the function to choose one item several times, but not the other way around.

grr:
> should the function that chooses multiple items choose with replacement? I suggested that it should not, as choosing multiple items with replacement can be reimplemented by calling the function to choose one item several times, but not the other way around.

I like how Perl6/Raku has two functions for the two behaviors: pick() and roll()
https://docs.perl6.org/type/List#routine_pick

leonerd:
The implementation is highly nonideal.

1) It uses a C99 VLA - good luck getting that through any of the less up-to-date compilers on exotic platforms.
2) The "redo" algorithm is going to slow down massively as the number of samples approaches the size of the list.

A far better algorithm is to

    n times:
        select one item at random from the list; push it to the result
        if you selected any element but the final element:
            move the final element of the list into the position you just selected
        shorten the list by one element

Given the thoughts on https://rt.cpan.org/Ticket/Display.html?id=131535 about a customisable rand() choice, I'll probably rewrite this.

grr:
> The implementation is highly nonideal.

> 2. The "redo" algorithm is going to slow down massively as the number of samples approaches the size of the list.

that was already pointed earlier

> A far better algorithm is to
> 
> ```
> n times:
>     select one item at random from the list; push it to the result
>     if you selected any element but the final element:
>         move the final element of the list into the position you just selected
>     shorten the list by one element
> ```

how is that better that the one i proposed and @y implemented?

leonerd:
This is now handled by #86 and released in 1.54

PR #44, opened 2016-05-28
support for utf8 and binary subroutine names and packages

rurban:
Add support for utf8 and binary subroutine names and packages.
Taken from https://github.com/rurban/sub-name/commits/binary

1.45_01: C++ and MSVC compilation patch

```
From  https://rt.cpan.org/Ticket/Display.html?id=113180
failed under C++ and older MSVC
```

Some 5.6 test fixes
skip utf8::encode and stringify overload

Various minor pod fixes.


rurban:
superceded by #52

PR #46, opened 2016-06-06
Polyfill for __SUB__ token

jberger:
Per the discussion on https://rt.cpan.org/Public/Bug/Display.html?id=111466 this PR implements a polyfill for the __SUB__ token. The code is mostly lifted from Sub::Current (Rafael Garcia-Suarez). I did the porting and Aaron Crane helped with an issue I ran into.

The one remaining question in my mind is if ROUTINE should be exported/exportable or not. The case for it is that it would allow a cautious user to always use the polyfill implementation rather than relying on __SUB__ being the polyfill on old perl versions vs being the core token in newer versions. The case against is the added complexity of having two different names doing the ostensibly the same thing. Thoughts?


lib/Sub/Util.pm
```diff
@@ -21,6 +21,28 @@ $VERSION   = eval $VERSION;
 require List::Util; # as it has the XS
 List::Util->VERSION( $VERSION ); # Ensure we got the right XS version (RT#100863)
 
+use constant CAN_SUB => $^V >= 5.016;
```
haarg: Should check against $], not $^V.

lib/Sub/Util.pm
```diff
@@ -21,6 +21,28 @@ $VERSION   = eval $VERSION;
...
+  *__SUB__ = \&ROUTINE;
+}
+
+sub import {
+  my ($class, @args) = @_;
+  if (CAN_SUB) {
+    my @new_args = grep { $_ ne '__SUB__' } @args;
+    if (@args != @new_args) {
+      feature->import('current_sub');
+      @_ = ($class, @new_args);
```
haarg: Modifying `@_` and then using goto causes segfaults on perl 5.8.4.

leonerd: Eh; so we only do this for 5.8.5 and above and tell older users "sorry, no can do"

haarg: Actually it doesn't matter - this code path is only for 5.10+.  My comment can be ignored.

jberger:
as discussed on IRC this PR should probably be held until ROUTINE is made private (I think that is the side we fell to) and some documentation is added.

lib/Sub/Util.pm
atoomic: `$^V` is not available in 5.6 for example and stands for `v5.x` in later version
even if the compare is not incorrect for modern Perl version
you should use `$]`

jberger:
Is it as simple as changing the variable? Do I also have to change the comparison?

Grinnz:
No, it is already comparing against the right decimal number.

jberger:
This branch is very stale but locally a rebased version passes tests. Any opposition to me force-pushing the rebased version?

atoomic:
I would say rebase && force push. The CI will confirm how it stands with the various Perl version
Note: we are currently using Travis CI, but I've provided a quick and fast alternate addition using GH CI as #110

jberger:
OK so it fails on 5.6. Is that fixable? Is it allowable?

atoomic:
@jberger the problem comes from `Perl_find_runcv`, when running ppport.h
we can see that this is not portable before 5.8.1
```
*** WARNING: Uses find_runcv, which may not be portable below perl 5.8.1, even with 'ppport.h'
```

So I came with a poor man alternate with an additional commit to provide it when missing, and it seems to work...
the alternate is to drop the support below 5.6

view commit:
https://github.com/atoomic/Scalar-List-Utils/commit/c74c017ef1b0e57a9af89b62bd42a5b61f8e750d

view CI proof of concept: 
https://github.com/atoomic/Scalar-List-Utils/actions/runs/205763453

So cherry picking this commit to your branch should fix your concern.

jberger:
> @jberger the problem comes from `Perl_find_runcv`, when running ppport.h
> we can see that this is not portable before 5.8.1
> [SNIP]
> So I came with a poor man alternate with an additional commit to provide it when missing, and it seems to work...
> the alternate is to drop the support below 5.6
> [SNIP]
> So cherry picking this commit to your branch should fix your concern.

That's beyond my power to evaluate. @leonerd thoughts?

PR #47, opened 2016-06-10
make the XS code ppport.h free when in core

bulk88:
ppport.h is an empty file on the Win32 Perl build and therefore
PERL_BCDVERSION macro doesn't exist/not defined. Use the core VERSION
macros, not the ppport family of VERSION macros. Make SLU ppport-free
when inside core.

### DIFF
--- a/ListUtil.xs
+++ b/ListUtil.xs
@@ -7,14 +7,33 @@
 #include <perl.h>
 #include <XSUB.h>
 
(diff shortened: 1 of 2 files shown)


bulk88:
Bump

leonerd:
There's a conflict in this

leonerd:
Merged. Maybe. possibly. Probably.

PR #48, opened 2016-07-17
Document the behaviour of uniqnum() on NaN

arc:
This PR merely documents the current behaviour (which treats all `NaN` values as equal, because it's using a hash to detect uniqueness of string representation).

It's possible that some people would prefer a strict IEEE approach of treating any `NaN` value as unequal to all other `NaN` values (including itself); or even an approach that treats `NaN` values as unequal iff their payloads differ. My guess is that the current behaviour offers least surprise for most people.


leonerd:
LGTM

PR #49, opened 2016-09-13
Patches from split

haarg:
These are various patches I wrote or took from other sources when attempting to split the modules into separate dists.


dur-randir:
Sub::Util part carries on all the problems accumulated around it:

- these don't compile before 5.16
- these fail some tests before 5.16
- these do not fix \0 mishandling
- these are broken under debugger

haarg:
Updating ppport.h fixes the issues with <5.16.  This patchset shouldn't cause any additional issues with \0 in sub names, so that can be addressed separately.

@dur-randir Can you elaborate on the issues with the debugger?

dur-randir:
>This patchset shouldn't cause any additional issues with \0 in sub names
If the intention is not a full binary/utf8 support here, then the commit "Support binary data and unicode in symbol names" has misleading name.

>Can you elaborate on the issues with the debugger?
%DB::sub entries are not correctly set for the same cases.

haarg:
What "same cases"?

dur-randir:
>What "same cases"?

Setting binary/utf8 name for a function.

haarg:
Re-reviewing the patches, they do handle \0 in sub names.  %DB::sub does have some edge cases though.

haarg:
I've updated this to fix the `%DB::sub` issues and included some other patches from Sub::Name.

ribasushi:
@haarg out of curiosity does the above address @chipdude's forgotten issue: https://rt.cpan.org/Ticket/Display.html?id=65540#txn-893983

haarg:
Split into #63, #64, #65, #66, #67

PR #52, opened 2016-12-14
support cperl, binary names, lexical $_

rurban:
* support PERL_HAS_QUOTE_PKGSEPERATOR (removed in v5.25.2c)
* support PERL_STRIP_NUL (<5.16 and >= v5.25.2c)
* support lexical $_ (find_rundefsvoffset)
* support binary and unicode names in subname
* Fix -Wc++11-compat warnings
* more t/subname.t unicode fixups for v5.24c and <5.16


karenetheridge:
Some of these commits use versions that have already been used in unauthorized releases. Other commits contain changes from the Sub-Name release, which should be copied with attribution and references.

t/subname.t
```diff
@@ -55,8 +55,7 @@ sub caller3_ok {
     # this is apparently how things worked before 5.16
     utf8::encode($expected) if $] < 5.016 and $ord > 255;
     # before 5.16 and after v5.25.2c names with NUL are stripped
-    my $strip_nul = ($] < 5.016 or ($Config{usecperl} and $] >= 5.025002));
-    if (!$ord and $strip_nul) {
+    if (!$ord and ($] < 5.016 or ($Config{usecperl} and $] >= 5.025002))) {
```
karenetheridge: These changes appear out of sync with the Sub-Name distribution (the corresponding file there is t/exotic_names.t).  It would be appreciated if the changes go there first, with suitable documentation that we understand what the problem(s) are.

ppport.h
```diff
@@ -4,9 +4,9 @@
 /*
 ----------------------------------------------------------------------
 
-    ppport.h -- Perl/Pollution/Portability Version 3.19_02
+    ppport.h -- Perl/Pollution/Portability Version 3.35_02
```
karenetheridge: ppport.h updates should occur in their own commit.

Where is 3.35_02 coming from? cpan only has 3.35. We should avoid using alpha releases unless absolutely necessary.

rurban:
> On Dec 14, 2016, at 7:32 PM, Karen Etheridge <notifications@github.com> wrote:
> 
> @karenetheridge commented on this pull request.
> 
> In ppport.h:
> 
> > @@ -4,9 +4,9 @@
>  /*
>  ----------------------------------------------------------------------
>  
> -    ppport.h -- Perl/Pollution/Portability Version 3.19_02
> +    ppport.h -- Perl/Pollution/Portability Version 3.35_02
> 
> ppport.h updates should occur in their own commit.
> 
> Where is 3.35_02 coming from? cpan only has 3.35. We should avoid using alpha releases unless absolutely necessary.

cperl, as mentioned in the subject line.
perl5 is not able to support binary symbols yet.

Reini Urban
rurban@cpan.org

karenetheridge:
> cperl, as mentioned in the subject line. perl5 is not able to support binary symbols yet.

I think you should get those changes into Devel::PPPort first, rather than putting them directly into ppport.h in this distribution.

karenetheridge:
"This branch has conflicts that must be resolved"

I suspect you have been using this branch for other purposes, and do not intend for this to be a pull request any longer?

rurban:
I do maintain my own fork with these features and fixes. If someone cares.

karenetheridge:
I'm sure more patches would be happily received -- but this particular PR and branch are too cluttered with conflicts to be the place to submit them.

PR #58, opened 2017-12-20
Fix indentation

majlo:
Nesting level doesn't match indentation and makes feel that "sv_2mortal"
is part of "if (spill)" branch, while it isn't.

For rt123781, done as part of CPAN Pull Request Challenge.

### DIFF
--- a/ListUtil.xs
+++ b/ListUtil.xs
@@ -952,11 +952,13 @@ PPCODE:
                     stack[reti++] = newSVsv(PL_stack_base[i + 1]);
         }
 
-        if (spill)
+        if (spill) {
             /* the POP_MULTICALL will trigger the SAVEFREESV above;
              * keep it alive  it on the temps stack instead */
             SvREFCNT_inc_simple_void_NN(spill);
-            sv_2mortal((SV*)spill);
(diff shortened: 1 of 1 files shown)


ListUtil.xs
```diff
@@ -952,11 +952,13 @@ PPCODE:
...
 
-        if (spill)
+        if (spill) {
             /* the POP_MULTICALL will trigger the SAVEFREESV above;
              * keep it alive  it on the temps stack instead */
             SvREFCNT_inc_simple_void_NN(spill);
-            sv_2mortal((SV*)spill);
+        }
+
+        sv_2mortal((SV*)spill);
```
leonerd: This looks wrong. I believe the original logic intention was that `sv_2mortal()` also be conditional on whether `spill` is true. As it happens, that line was executed unconditionally, and apparently that it doesn't crash all the time is simply because `sv_2mortal()` is apparently OK to be called on NULL. But nevertheless, the call ought to be inside the conditional braces.

leonerd:
I won't merge this PR itself, but I've fixed the actual braces issue in 0f0539f

PR #59, opened 2017-12-20
Add appveyor config for testing on Windows

lancew:
This commit is done as part of the CPAN Pull Request Challenge 2017.
Basic appveyor config to test that the module build on Windows.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,0 +1,15 @@
+cache:
+  - C:\strawberry
+
+install:
+  - if not exist "C:\strawberry" choco install strawberryperl -y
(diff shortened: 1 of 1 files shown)


karenetheridge:
What is needed to enable appveyor builds? Or will they happen automatically given the presence of the config file?

lancew:
Hi, as I understand it... I believe someone would need to register/sign-in on Appveyor (https://www.appveyor.com/) using GitHub credentials and then on https://ci.appveyor.com/projects/new add the project.

Then the config in this PR would be used.

tonycoz:
This is probably obsolete, SLU now uses github actions to test against strawberry perl.

PR #60, opened 2017-12-20
Rt #122582

colinnewell:
This is the simplest possible fix for the test case described in [RT #122582](https://rt.cpan.org/Ticket/Display.html?id=122582).  This just ensures that the IV is copied over as well as the NV if present.

My knowledge of Perl internals isn't sufficient to really know if this is sufficient to fix the full bug described by zefram.


t/dualvar.t
```diff
@@ -131,3 +131,6 @@ SKIP: {
   ok(isdual($ary[2]), 'Is a dualvar');
 }
 
+# RT #122582: dual var not copying numeric value correctly
+my $a = "4611686018427387915";
```
leonerd: This test is only going to pass correctly on 64bit perls surely?

ListUtil.xs
```diff
@@ -1200,7 +1200,7 @@ CODE:
         SvIsUV_on(TARG);
     }
 #endif
-    else {
+    if(SvIOK(num)) {
```
leonerd: If this `else` is going to disappear, I feel that the `else` on line 1197 above ought to too.

colinnewell:
Just figured I should say I will take a look at this.  I've just struggled to setup a 32 bit test perl without 64 bit int support easily so I haven't actually done anything with this yet.

colinnewell:
Sorry it's taken so long, but I've rebased and updated based on your feedback.  I think the Travis fail is just a glitch, if you hit rebuilt it will probably work..

PR #61, opened 2017-12-31
 Implement some functions as custom OPs

JRaspass:
Someone with some XS knowledge should probably check this :-P And we probably need to add some threaded perls to travis for good measure, but this commit makes a few of the simple Scalar::Util functions custom OPs on 5.14+.

The work was **heavily** based on Ref::Util, so thanks @xsawyerx :+1: 

@leonerd You mentioned you wanted custom OPs for List::Util back in a blog post in 2014 - http://leonerds-code.blogspot.co.uk/2014/06/listutil-additions-in-perl-520.html and while this doesn't get us to the List::Util functions yet, it's a start.

This commit is done as part of the CPAN Pull Request Challenge 2017.


karenetheridge:
that's not the newest ppport.h :)  I've committed the newest one to master.

JRaspass:
Rebased without the ppport.h change, thanks @karenetheridge. I'm happy to rebase again once #62 is merged so we have some confidence on threaded perls.

dur-randir:
This looks fine to be run on threaded perls. If you want to stress-test it, make a test that does threads->create(sub {require S::U; S::U::somefunc();}) in a loop.

leonerd:
Hmm. That's quite a lot of change to turn up in one PR unannounced and (at least to me) undiscussed. I'll need a bit of time to read this over and see what I think to it.

leonerd:
Oops - didn't mean to hit the 'close' button

JRaspass:
@leonerd Sorry it was undiscussed, S-L-U was my company's module for Dec's CPAN Pull Request Challenge so this PR was more or less dreamt up and executed within a night. So on that basis feel free to abandon it if you feel it's too big a change, it was more a learning experience than anything else.

leonerd:
@JRaspass Oh quite the opposite. If I didn't like the look of it at all indeed I would have closed it. I felt it polite to let you know "I'm looking/thinking on it", to excuse my lack of resolution one way or the other.

The eventual goal of this PR is definitely a worthy aim, and something I'm keen to see. My hesitation was just that it's a lot of code to implement it, and I need to carefully consider *how* it's achieved, and see if that looks the way we want to use.

leonerd:
Yes, I think actually that's looking quite nice.

One additional thing you can consider doing - making the appropriate fixes to `B::Deparse` to make deparsing around these happy. That's actually a fairly simple one, because `B::Deparse` invokes a method on itself dynamically by name based on the opname of each op in the tree it encounters. Because your new ops here all have new names, it's easy to name those.

Also I'd be interested to see some benchmark comparisons, both of tiny "one-line" functions and maybe find some more typical use-cases from CPAN of larger functions including theses ops in them, to see what realworld improvement they actually make.

JRaspass:
Thanks for the feedback Paul, I've amended the PR to include the B::Deparse changes (https://github.com/Dual-Life/Scalar-List-Utils/pull/61/files#diff-64290629be4caf8f7d4dab481f42177b) but I wanted your feedback on wether this is the right approach since I couldn't actually find any documentation on how to do this other than your p5p post (https://groups.google.com/d/msg/perl.perl5.porters/bEz655Ocaeg/ISixn1F_dbMJ) and the source of B::Tap (https://metacpan.org/source/TOKUHIROM/B-Tap-0.15/lib/B/Tap.pm).

I'll try and get to benchmarks this week, custom op vs $sub->() should be easy enough to compare but I presume $sub->() is already slower than sub(). Or I could temporarily put clones of the XS functions in without OPs under new names, shout if you have any good ideas on benching them.

JRaspass:
I guess I should probably add a deparse.t now, /me adds that to the todo list :-P

atoomic:
@leonerd at first glance these optimizations using custom OPs seems interesting
What's the status on this change?

leonerd:
@atoomic Honestly, lack of time at this point. Between the distractions of "Perl 7" as a plan, that's eating into time I would be spending implementing `FINALLY`, `try/catch`, `Object::Pad` then thoughts of dumb-match, type systems, macros,... and that's just the "free software" things. I'm then doing libuv wrapping for TPF, and I actually have at least 2 actual day-jobs doing perl and not-perl things. In summary, there's only so many hours in the day ;)

atoomic:
@leonerd I think these are all valid reasons. I was just wandering why there was no activity since January 2018. Let me know if I can help.

JRaspass:
@leonerd Shall I close this PR now in favour of your [List::Keywords](https://metacpan.org/pod/List::Keywords) dist? I presume there will be a `Scalar::Keywords` at some point too and maybe one day have all the keywords in core :-P

Also is there a repo for the dist yet? I wanted to fix a small bug in the rendered POD where it shows `none =head2 notall`

leonerd:
> @leonerd Shall I close this PR now in favour of your [List::Keywords](https://metacpan.org/pod/List::Keywords) dist? 

Yeah could well do. But see below.

> I presume there will be a `Scalar::Keywords` at some point too and maybe one day have all the keywords in core :-P

That one is far less interesting. The scalar utils all happen to be function-like keywords, so to be honest those can be inlined into real OPs transparently. There's a case for still keeping that here.

> Also is there a repo for the dist yet? I wanted to fix a small bug in the rendered POD where it shows `none =head2 notall`

I've already fixed in source. :)

leonerd:
In light of the fact that core perl provides many of these functions as possibly custom ops since 5.36, and that Scalar::Util itself will transparently just use those on 5.40+ where they are now stable, I think most of this PR is now redundant and can be closed. In any case, it currently has conflicts.

@JRaspass Do you want to take a look, or can we just close it now?

JRaspass:
Agreed, the new builtin:: approach is much nicer for handling most of Scalar::Util. And hopefully things like https://github.com/Perl/PPCs/pull/50 go through for handing List::Util. Closing this :+1:

PR #63, opened 2018-01-03
fix subname when stash is deleted

haarg:


karenetheridge:
trivial typo in commit message: "What gets lots.."

t/subname.t
```diff
@@ -82,12 +82,13 @@ is($x->(), "main::foo");
...
   {
     package Blarf;
-    sub gorp { (caller(0))[3] }
+    sub gorp { 1 }
   }
   my $sub = \&Blarf::gorp;
   delete $::{'Blarf::'};
-  my $name = $sub->();
-  is subname($sub), $name,
+  my $package = $] < 5.010 ? 'Blarf' : '__ANON__';
```
karenetheridge: Zefram was recommending that all uses of $] should use string interpolation to avoid floating point issues: e.g. `my $package = "$]" < '5.010' ..`

haarg:
Addressed @karenetheridge's comments.

haarg:
This was fixed by 2e24b7c78649188ba4d2eb1166620b25140fd4e4.

PR #66, opened 2018-01-03
test reduce aliasing arguments

haarg:


leonerd:
I don't want to assert this in the tests because the documentation never states it. What it says is:

> Reduces @list by calling "BLOCK" in a scalar context multiple times, setting $a and $b each time. The first call will be with $a and $b set to the first two elements of the list, subsequent calls will be done by setting $a to the result of the previous call and $b to the next element in the list.

It just uses the phrase "sets $b to ...", doesn't mention aliasing. The current implementation happens to alias but I don't want to guarantee that in future, as that's a fragile implementation detail that might change.

jjatria:
If this is not going to be merged, should this be closed instead?

PR #67, opened 2018-01-03
Fix set_subname with exotic names

haarg:
This includes unicode and binary names.  Patches are primarily taken from Sub::Name.

### DIFF
--- a/ListUtil.xs
+++ b/ListUtil.xs
@@ -1485,7 +1485,8 @@ PREINIT:
     STRLEN namelen;
     const char* nameptr = SvPV(name, namelen);
     int utf8flag = SvUTF8(name);
-    int seen_quote = 0, need_subst = 0;
+    int quotes_seen = 0;
+    bool need_subst = FALSE;
 PPCODE:
     if (!SvROK(sub) && SvGMAGICAL(sub))
         mg_get(sub);
@@ -1508,21 +1509,21 @@ PPCODE:
         if (s > nameptr && *s == ':' && s[-1] == ':') {
             end = s - 1;
             begin = ++s;
-            if (seen_quote)
-                need_subst++;
+            if (quotes_seen)
+                need_subst = TRUE;
         }
         else if (s > nameptr && *s != '\0' && s[-1] == '\'') {
             end = s - 1;
             begin = s;
-            if (seen_quote++)
-                need_subst++;
+            if (quotes_seen++)
+                need_subst = TRUE;
         }
     }
     s--;
     if (end) {
         SV* tmp;
         if (need_subst) {
-            STRLEN length = end - nameptr + seen_quote - (*end == '\'' ? 1 : 0);
+            STRLEN length = end - nameptr + quotes_seen - (*end == '\'' ? 1 : 0);
             char* left;
             int i, j;
             tmp = newSV(length);
--- a/ListUtil.xs
+++ b/ListUtil.xs
@@ -1487,6 +1487,7 @@ PPCODE:
         namelen -= end - nameptr;
     }
 
+    #ifdef PERL_VERSION < 10
     /* under debugger, provide information about sub location */
     if (PL_DBsub && CvGV(cv)) {
         HV *hv = GvHV(PL_DBsub);
@@ -1521,6 +1522,7 @@ PPCODE:
         }
         Safefree(full_name);
     }
+    #endif
 
     gv = (GV *) newSV(0);
     gv_init_pvn(gv, stash, nameptr, s - nameptr, GV_ADDMULTI | utf8flag);
--- a/ListUtil.xs
+++ b/ListUtil.xs
@@ -1454,8 +1454,8 @@ PREINIT:
     const char *s, *end = NULL, *begin = NULL;
     MAGIC *mg;
     STRLEN namelen;
-    int utf8flag = SvUTF8(name);
     const char* nameptr = SvPV(name, namelen);
+    int utf8flag = SvUTF8(name);
     int seen_quote = 0, need_subst = 0;
 PPCODE:
     if (!SvROK(sub) && SvGMAGICAL(sub))
--- a/ListUtil.xs
+++ b/ListUtil.xs
@@ -1451,11 +1451,12 @@ PREINIT:
     CV *cv = NULL;
     GV *gv;
     HV *stash = CopSTASH(PL_curcop);
-    const char *s, *end = NULL;
+    const char *s, *end = NULL, *begin = NULL;
     MAGIC *mg;
     STRLEN namelen;
     int utf8flag = SvUTF8(name);
     const char* nameptr = SvPV(name, namelen);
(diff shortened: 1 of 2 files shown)


karenetheridge:
+1

With these changes I would feel more confident in switching Moose from Sub::Name to Sub::Util (see https://github.com/moose/Moose/pull/127)

dur-randir:
Is this any different from #49?

haarg:
I split some of the commits out into separate PRs.

ListUtil.xs
```diff
@@ -1445,14 +1474,18 @@ PPCODE:
...
     CV *cv = NULL;
     GV *gv;
     HV *stash = CopSTASH(PL_curcop);
-    char *s, *end = NULL;
+    const char *s, *end = NULL, *begin = NULL;
     MAGIC *mg;
+    STRLEN namelen;
+    const char* nameptr = SvPV(name, namelen);
+    int utf8flag = SvUTF8(name);
+    int seen_quote = 0, need_subst = 0;
```
leonerd: I'm not terribly comfortable with the bookkeeping of `need_subst` and `seen_quote` added in here.

Both variables are declared as `int` types, with names that feel booleanish, but then one of them is used as a counter anyway.

Can you declare these as

    int quotes_seen = 0;
    bool need_subst = FALSE;

Then the various places that `need_subst++` should become

    need_subst = TRUE;

ListUtil.xs
```diff
@@ -1465,63 +1498,81 @@ PPCODE:
...
-        if (*s == ':' && s[-1] == ':')
-            end = ++s;
-        else if (*s && s[-1] == '\'')
-            end = s;
+    for (s = nameptr; s <= nameptr + namelen; s++) {
+        if (s > nameptr && *s == ':' && s[-1] == ':') {
+            end = s - 1;
+            begin = ++s;
+            if (seen_quote)
+                need_subst++;
```
leonerd: Why is this line needed? `need_subst` is only used as its truthy nature (see comment above about declaring it `bool`), and the only occasion in which `seen_quote` would be true would have already set `need_subst` true anyway.

ListUtil.xs
```diff
@@ -1465,63 +1498,81 @@ PPCODE:
...
-        char *namepv = savepvn(name, end - name);
-        stash = GvHV(gv_fetchpv(namepv, TRUE, SVt_PVHV));
-        Safefree(namepv);
-        name = end;
+        SV* tmp;
+        if (need_subst) {
+            STRLEN length = end - nameptr + seen_quote - (*end == '\'' ? 1 : 0);
+            char* left;
+            int i, j;
+            tmp = newSV(length);
```
leonerd: tmp = sv_2mortal(newSV(length));

would mean you don't have to remember to dec it later

ListUtil.xs
```diff
@@ -1465,63 +1498,81 @@ PPCODE:
...
+                if (nameptr[j] == '\'') {
+                    left[i] = ':';
+                    left[++i] = ':';
+                }
+                else {
+                    left[i] = nameptr[j];
+                }
+            }
+            stash = gv_stashpvn(left, length, GV_ADD | utf8flag);
+            SvREFCNT_dec(tmp);
```
leonerd: Unnecessary if you mortalise it earlier

ListUtil.xs
```diff
@@ -1465,63 +1498,81 @@ PPCODE:
...
+        if (s > nameptr && *s == ':' && s[-1] == ':') {
+            end = s - 1;
+            begin = ++s;
+            if (seen_quote)
+                need_subst++;
+        }
+        else if (s > nameptr && *s != '\0' && s[-1] == '\'') {
+            end = s - 1;
+            begin = s;
+            if (seen_quote++)
```
leonerd: Are you quite sure you mean `seen_quote++` and not `++seen_quote` here? I.e. the very first time you see a quote mark you increment `seen_quote` but don't yet set `need_subst` to true, until the second time you see one?

ListUtil.xs
```diff
@@ -1465,63 +1498,81 @@ PPCODE:
...
+        SV* old_full_name = newSVpvn_flags(HvNAME(oldhv), HvNAMELEN_get(oldhv), HvNAMEUTF8(oldhv) ? SVf_UTF8 : 0);
+        sv_catpvn(old_full_name, "::", 2);
+        sv_catpvn_flags(old_full_name, GvNAME(oldgv), GvNAMELEN(oldgv), GvNAMEUTF8(oldgv) ? SV_CATUTF8 : SV_CATBYTES);
+
+        old_data = hv_fetch_ent(DBsub, old_full_name, 0, 0);
+
+        SvREFCNT_dec(old_full_name);
+
+        if (old_data && HeVAL(old_data)) {
+            SV* new_full_name = newSVpvn_flags(HvNAME(stash), HvNAMELEN_get(stash), HvNAMEUTF8(stash) ? SVf_UTF8 : 0);
```
leonerd: Ditto comment about mortalising here

ListUtil.xs
haarg: If only a single quote character is seen, and no `::` package separator, then the quote is separating the package name from the sub name.  In that case, we don't need to construct a new package name using double colons (which is what needs_substr tracks) because the package name won't have any single quotes (or `::`) in it.  This also explains why the seen_quote check exists in the branch for `::`.

ListUtil.xs
Leont: That's probably a historical accident (it's been a while since I touched this, but I think that in an earlier version the two were one variable).

ListUtil.xs
leonerd: Ah, I see. In that case, maybe rather than two variables could we just count the number of quotes seen (named `quotes_seen` so a plural noun suggesting a count), and change the condition to something like

    if(quotes_seen > 1) {
        /* multiple quotes were seen, but only the final one separates the basename from the package.
           the rest need converting into :: */

Leont: It matters where the quote is seen. `Foo::Bar'baz` is fine, but `Foo'Bar::baz` isn't.

Parsing in C sucks.

ListUtil.xs
Leont: I'd call that a stylistic difference. I don't care much either way, but lets' please keep Sub::Util and Sub::Name in sync :-)

haarg:
@leonerd I've addressed your comments.

karenetheridge:
Do any of the changes here need to get propagated back to Sub::Name?

haarg:
There is no difference in behavior, but if this gets merged I'll prepare a patch for Sub::Name to bring them in line.

Leont:
> if this gets merged I'll prepare a patch for Sub::Name to bring them in line.

@haarg++

leonerd:
It's still a bit messy in places, but I guess that's to be expected of this low level of name mangling. For sake of moving it forward (and that it's consistent with `Sub::Name` anyway), LGTM

PR #68, opened 2018-01-20
Fix tests to work also on Perl 5.6.2

pali:
It is better to run tests instead of skipping them.


leonerd:
Seems good. Shall wait for travis to finish

pali:
Travis has no problems.

PR #70, opened 2018-02-25
optimize head/tail

dur-randir:
Seeing e559254d839cae38de3013455250680e10d190dd, I couldn't pass. This greatly speeds it up and cuts code to half from the original.

### DIFF
--- a/ListUtil.xs
+++ b/ListUtil.xs
@@ -623,35 +623,21 @@ PPCODE:
 {
     int size = 0;
     int start = 0;
-    int end = 0;
-    int i = 0;
 
-    size = SvIV(*(SP + 1));
+    if (items > 1) {
+        size = SvIV(*(SP + 1));
 
-    if ( ix == 0 ) {
-        start = 1;
-        end = start + size;
-        if ( size < 0 ) {
-            end += items - 1;
-        }
-        if ( end > items ) {
-            end = items;
-        }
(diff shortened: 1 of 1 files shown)


ListUtil.xs
```diff
@@ -623,44 +623,26 @@ PPCODE:
...
-    }
-    else {
-        EXTEND( SP, end - start );
-        for ( i = start; i <= end; i++ ) {
-            PUSHs( sv_2mortal( newSVsv( ST(i) ) ) );
-        }
-        XSRETURN( end - start );
+    while (--size >= 0) {
+        ++SP;
+        *SP = *(SP + start);
```
leonerd: Hmm.. This is quite the change of behaviour - we're now returning aliases of the original input list, rather than new temporary SVs that copy it. This would have a change of behaviour in lvalue context; for example

    my @n = (1 .. 10);
    $_++ for tail 4, @n;

Grinnz: If such code wouldn't work before, I'd say this would just be a new feature...

leonerd: It could be considered a feature yes. But it's a somewhat surprising change of behaviour. Is there a chance of accidental collateral damage?

haarg: Returning aliases is the behavior I would expect.

PR #71, opened 2018-04-03
Fix uniqnum for large floats and numeric strings

ilmari:
- Call SvNV on scalars with no numeric flags set (this sets SVf_IOK/SVf_UOK as appropriate)
- Use 15 digits of precision for NVs (matches Perl's default stringification)


ilmari:
A previous rebase appears to have lost the actual change for this PR, fixed.

leonerd:
LGTM

PR #74, opened 2018-11-22
Programming more defensively to avoid SEGV

mohawk2:
In certain circumstances, doing a `set_subname` has caused SEGV while running under Devel::Cover. I have not dug very deeply into exactly why, but I can show the repro case if needed. The immediate reason is that in that scenario, `GvSTASH(oldgv)` returns NULL, yet HvNAME is called on it.

This patch makes the block only continue with the attempt to "under debugger, provide information about sub location" if `oldhv` is not NULL.

### DIFF
--- a/ListUtil.xs
+++ b/ListUtil.xs
@@ -1605,19 +1605,21 @@ PPCODE:
 
         GV* oldgv = CvGV(cv);
         HV* oldhv = GvSTASH(oldgv);
-        SV* old_full_name = sv_2mortal(newSVpvn_flags(HvNAME(oldhv), HvNAMELEN_get(oldhv), HvNAMEUTF8(oldhv) ? SVf_UTF8 : 0));
-        sv_catpvn(old_full_name, "::", 2);
-        sv_catpvn_flags(old_full_name, GvNAME(oldgv), GvNAMELEN(oldgv), GvNAMEUTF8(oldgv) ? SV_CATUTF8 : SV_CATBYTES);
+        if (oldhv) {
+            SV* old_full_name = sv_2mortal(newSVpvn_flags(HvNAME(oldhv), HvNAMELEN_get(oldhv), HvNAMEUTF8(oldhv) ? SVf_UTF8 : 0));
+            sv_catpvn(old_full_name, "::", 2);
+            sv_catpvn_flags(old_full_name, GvNAME(oldgv), GvNAMELEN(oldgv), GvNAMEUTF8(oldgv) ? SV_CATUTF8 : SV_CATBYTES);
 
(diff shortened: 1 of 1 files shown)


karenetheridge:
This should also be ported to Sub::Name as well.

leonerd:
(as per IRC but here again for completeness) Can you provide a test case, or otherwise an example of what situations cause the bug that this code is fixing?

mohawk2:
Minimal repro:
- `git clone git@github.com:preaction/Yancy.git`
- Put the text below in `t/standalone-segv.t`
- `perl -MDevel::Cover -Ilib t/standalone-segv.t`

```
use Test::Mojo;
use FindBin qw( $Bin );
use Mojo::File qw( path );
use lib "".path( $Bin, 'lib' );
use Local::Test qw( init_backend );

my $collections = {};
my ( $backend_url, $backend, %items ) = init_backend($collections);

my $t = Test::Mojo->new( 'Yancy', {
    backend => $backend_url,
    collections => $collections,
    read_schema => 1,
} );

$t->get_ok( '/' )
  ->status_is( 404 )
  ->text_is( h1 => 'Welcome to Yancy' )
  ;

$t = Test::Mojo->new( 'Yancy', {
    backend => $backend_url,
    collections => $collections,
} );
$t->get_ok( '/yancy' )
  ->status_is( 200 )
  ;
```

mohawk2:
Actually, due to lack of functioning email on my Linux box, I'll have to leave that porting to someone else. Hopefully not a big ask :-)

rurban:
This fix is wrong.
I fixed this in cperl by checking for !oldhv or oldhv == PL_defstash, and then I'm using a new empty string for the stash name.
Same problem exists for the new stash also.

mohawk2:
@Leont belatedly:

```
#0  XS_Sub__Util_set_subname (cv=<optimised out>) at ListUtil.xs:1608
#1  0x000055f495ea5dd6 in Perl_pp_entersub () at pp_hot.c:5232
#2  0x000055f495e61c52 in Perl_runops_debug () at dump.c:2536
#3  0x000055f495dd535e in S_run_body (oldscope=1) at perl.c:2689
#4  perl_run (my_perl=<optimised out>) at perl.c:2617
#5  0x000055f495d9eda2 in main (argc=<optimised out>, argv=<optimised out>, 
    env=<optimised out>) at perlmain.c:122
```

Do you want me to change my fix to be the same as @rurban 's? Or can you just cherry-pick his? I'm keen for SEGVs to not happen, I don't mind whose code makes it so :-)

Leont:
In either case, this really needs a test.

mohawk2:
It certainly does. Sadly I don't know what triggers the behaviour it's protecting from.

PR #77, opened 2019-08-02
Fix segfaults in (set_)subname with deleted stashes

ilmari:
Return `__ANON__` as the package name if the stash has been deleted.

### DIFF
--- a/ListUtil.xs
+++ b/ListUtil.xs
@@ -1601,15 +1601,18 @@ PPCODE:
     /* under debugger, provide information about sub location */
     if (PL_DBsub && CvGV(cv)) {
         HV* DBsub = GvHV(PL_DBsub);
-        HE* old_data;
+        HE* old_data = NULL;
 
         GV* oldgv = CvGV(cv);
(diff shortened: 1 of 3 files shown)


karenetheridge:
Is this fix also relevant to Sub::Name?

ilmari:
@karenetheridge Yes, both `Sub::Name` and `Sub::Identify` are affected by this.

leonerd:
I wonder if `__ANON__` would be at all confusing.. Is there any particular historical / back-compat reason why it has to be that string? Could something like `(deleted)` be used instead? That might be less confusing

ilmari:
Perl core's `gv_fullname4` function uses `__ANON__` when it can't find a stash: https://github.com/Perl/perl5/blob/blead/gv.c#L2490

leonerd:
Oh I see. A little unfortunate - I had hoped to retain "__ANON__" as a package name for present-but-unnamed packages, akin to anon subs.

Mmm well in that case this will just have to do the same.

jrubinator:
For my edification: when can/should I expect this fix to reach cpan?

PR #78, opened 2019-08-13
Fix uniqnum on large stringified integers on long- and quad-double perls

ilmari:
Turns out long-double perls' `SvNV()` macro doesn't set the integer
flags even if the value is an integer.  Use `SvIV()` instead, which
appears to be doing the right thing for normal, long and quad doubles.


leonerd:
LGTM

PR #79, opened 2019-08-31
Corrections to uniqnum

sisyphus:
The uniqnum() sub has some bugs that the current t/uniq.t test file fails to expose.
This pull request includes a version of t/uniq.t that has been expanded to reveal the failures of the current Scalar-List-Utils-1.52 distribution.

See https://rt.cpan.org/Public/Bug/Display.html?id=130302 for some discussion on this, and for a none-too-pretty view of the evolution of my patches.

Cheers,
Rob

### DIFF
--- a/ListUtil.xs
+++ b/ListUtil.xs
@@ -72,6 +72,19 @@
 #define sv_catpvn_flags(b,n,l,f) sv_catpvn(b,n,l)
 #endif
 
+#ifdef UNIQ_PREC_LOSS
+#define UNIQ_BUFF_SIZE 33                    /* nvsize == 8, ivsize == 8 */
+
+#elif UNIQ_LOOP_END == 8                     /* nvsize == 8, ivsize < 8 **/
+#define UNIQ_BUFF_SIZE 17
+
+#elif UNIQ_LOOP_END == 10                    /* NV is 10 bytes ***********/
+#define UNIQ_BUFF_SIZE 21
+
+#else
+#define UNIQ_BUFF_SIZE 33                    /* NV is 16 bytes ***********/
+#endif
+
 /* Some platforms have strict exports. And before 5.7.3 cxinc (or Perl_cxinc)
    was not exported. Therefore platforms like win32, VMS etc have problems
    so we redefine it here -- GMB
@@ -1152,6 +1165,15 @@ CODE:
     int index;
     SV **args = &PL_stack_base[ax];
     HV *seen;
+    size_t i;
+    NV nv_arg;
+    void *p = &nv_arg;
+    char s[UNIQ_BUFF_SIZE];
+    char * buff = s;
+#ifdef UNIQ_PREC_LOSS
+    size_t ivstring = 0, offset = 0;
+    IV int_arg;
+#endif
 
     if(items == 0 || (items == 1 && !SvGAMAGIC(args[0]) && SvOK(args[0]))) {
         /* Optimise for the case of the empty list or a defined nonmagic
@@ -1185,12 +1207,106 @@ CODE:
 #endif
             }
 
-            if(!SvOK(arg) || SvUOK(arg))
-                sv_setpvf(keysv, "%" UVuf, SvUV(arg));
-            else if(SvIOK(arg))
-                sv_setpvf(keysv, "%" IVdf, SvIV(arg));
-            else
-                sv_setpvf(keysv, "%.15" NVgf, SvNV(arg));
+            /* Assign appropriate value to nv_arg */
+
+            if(!SvOK(arg) || SvUOK(arg)) {
+                nv_arg = (NV)SvUV(arg);
+#ifdef UNIQ_PREC_LOSS
+                if(SvUV(arg) > 9007199254740992) { /* UV to NV conversion could lose precision */
+                    ivstring = 1;
+                    int_arg = SvIV(arg); /* SvIV(arg) and SvuV(arg) have the same byte structure *
+                                          * and it's only the byte structure that interests us   */
+                }
+#endif
+            }
+
+            else if(SvIOK(arg)) {
+                nv_arg = (NV)SvIV(arg);
+#ifdef UNIQ_PREC_LOSS
+                int_arg = SvIV(arg);
+                if(int_arg < -4503599627370496)  /* IV to NV conversion could lose precision */
+                    ivstring = 1;
+#endif
+            }
+
+            else {
+                nv_arg = SvNV(arg);
+            }
+
+            /* Handle NaN, zeros and Infs */
+
+            if(nv_arg != nv_arg) {
+                sv_setpvf(keysv, "%s", "NaN");
+#ifdef UNIQ_PREC_LOSS
+                ivstring = 0;
+#endif
+            }
+
+            else if(nv_arg == 0) {
+                sv_setpvf(keysv, "%s", "0");
+#ifdef UNIQ_PREC_LOSS
+                ivstring = 0;
+#endif
+            }
+
+            else if(nv_arg/nv_arg != 1) {
+                if(nv_arg < 0) sv_setpvf(keysv, "%s", "-Inf");
+                else sv_setpvf(keysv, "%s", "Inf");
+#ifdef UNIQ_PREC_LOSS
+                ivstring = 0;
+#endif
+            }
+
+            /* Handle all values other than NaN, zeros and Infs */
+
+            else {
+#ifdef UNIQ_PREC_LOSS
+                if(ivstring) { 
+
+                    /* Read the bytes of iv_arg into the string buffer, in hex */
+
+                    sprintf(buff, "%016" UVxf, int_arg);
+                    buff += 16;
+                    offset = 16;
+                }
+
+
+                /* Take appropriate action if nv_arg holds an integer value    *  
+                 * within the IV and UV range that can lose precision as an NV */
+
+                else if(ceil(nv_arg) == nv_arg
+                        && ((nv_arg  <  1.8446744073709552e+19 && nv_arg > 9007199254740992)
+                             ||
+                            (nv_arg < -4503599627370496 && nv_arg  >= -9.2233720368547758e+18)) ) {
+
+                    int_arg = SvIV(arg);
+
+                    /* Read the bytes of int_arg into the string buffer, in hex */
+
+                    sprintf(buff, "%016" UVxf, int_arg);
+                    buff += 16;
+                    offset = 16; /* the buff pointer has been incremented by 16 */
+                }
+
+                ivstring = 0;
+
+                for(i = 0; i < 8; i++) {
+                    sprintf(buff, "%02x", ((unsigned char*)p)[i]);
+                    buff += 2;
+                }
+
+                buff -= 16 + offset; /* return pointer to original position */
+                offset = 0;
+#else
+                for(i = 0; i < UNIQ_LOOP_END; i++) {
+                    sprintf(buff, "%02x", ((unsigned char*)p)[i]);
+                    buff += 2;
+                }
+
+                buff -= 2 * UNIQ_LOOP_END; /* return pointer to original position */
+#endif
+                sv_setpvf(keysv, "%s", buff);
+            }
 #ifdef HV_FETCH_EMPTY_HE
             he = (HE*) hv_common(seen, NULL, SvPVX(keysv), SvCUR(keysv), 0, HV_FETCH_LVALUE | HV_FETCH_EMPTY_HE, NULL, 0);
             if (HeVAL(he))
--- a/ListUtil.xs
+++ b/ListUtil.xs
@@ -1171,7 +1171,7 @@ CODE:
     char s[UNIQ_BUFF_SIZE];
     char * buff = s;
 #ifdef UNIQ_PREC_LOSS
-    size_t ivstring = 0, offset = 0;
+    size_t potential_prec_loss = 0, offset = 0;
     IV int_arg;
 #endif
 
@@ -1212,11 +1212,10 @@ CODE:
             if(!SvOK(arg) || SvUOK(arg)) {
                 nv_arg = (NV)SvUV(arg);
 #ifdef UNIQ_PREC_LOSS
-                if(SvUV(arg) > 9007199254740992) { /* UV to NV conversion could lose precision */
-                    ivstring = 1;
-                    int_arg = SvIV(arg); /* SvIV(arg) and SvuV(arg) have the same byte structure *
-                                          * and it's only the byte structure that interests us   */
-                }
+                int_arg = SvIV(arg); /* SvIV(arg) and SvuV(arg) have the same byte structure *
+                                      * and it's only the byte structure that interests us   */
+                if(SvUV(arg) > 9007199254740992)  /* UV to NV conversion could lose precision */
+                    potential_prec_loss = 1;
 #endif
             }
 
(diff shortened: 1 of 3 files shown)


sisyphus:
Just realized that t/uniq.t does *not* fail tests 16 & 17 (uniqnum preserves uniqness of high precision floats) for perls whose nvsize=8 && ivsize=8.
This is actually a mistake in the construction of those 2 tests - which look at the list returned by
uniqnum(1.4142135623730951, 1.4142135623730954, 9223372036854775808, 9.2233720368547758e+18

Correct result on such a build of perl would print out the list as being (1.4142135623731, 1.4142135623731, 9223372036854775808).
S-L-U-1.52 will print out the returned list as being (1.4142135623731, 9223372036854775808 9.22337203685478e+18).

Those tests pass with S-L-U-1.52 because only the size (not the contents) of the list is being looked at.

A corrected t/uniq.t, which moves the comparison of 9223372036854775808 and 9.2233720368547758e+18 to test 12 (uniqnum compares numbers) is attached:

[uniq.t.TXT](https://github.com/Dual-Life/Scalar-List-Utils/files/3563000/uniq.t.TXT)

What's the best way to make this correction to the pull request ? Do I just push this latest version of t/uniq.t to my forked Scalar-List-Utils and make a second pull request ?

Also, is there any validity to the failing Travis checks ?
They purport to be in relation to perl-5.6.2, but the log seems to claim a failure with 5.22.1.
I'm unable to find any such problem with my 5.22.1 build on Ubuntu-18.04

karenetheridge:
> What's the best way to make this correction to the pull request ? Do I just push this latest version of t/uniq.t to my forked Scalar-List-Utils and make a second pull request ?

Amend your original commit (`git commit --amend t/uniq.t`, then `git push --force`. The pull request will update automatically.

karenetheridge:
(earlier comment edited, in case you are following along by email)

sisyphus:
Thanks Karen - most helpful.
Does the 5.6.2 test failure need to be investigated ?
It looks to me that the failure occurs because, in the list that's passed to uniqnum(), the first 2 elements ($maxuint and $maxuint - 1) are being coerced to NVs, which does indeed make them equivalent.

I don't see how my proposed changes would be responsible for that coercion, and I'm therefore thinking that this failure pre-dates my patches.
But I don't have any proof of that, and 5.6.2 is not building out of the box for me on either Linux or Windows.

karenetheridge:
The master build passes -- https://travis-ci.org/Dual-Life/Scalar-List-Utils/builds/573198383 -- so there must be an explanation for this failure that is caused by your code change.

sisyphus:
Yes ... it also (eventually) occurred to me that I could check that.
I concur.
Looks like I might have to dig out one of my old machines, hook up the
peripherals, and fire it up. (I know of at least one box that's got
perl-5.6.2 on it.)

sisyphus:
I was unable to get access to a perl-5.6.2 on any of my old machines -
except for a Windows laptop where the perl-5.6.2 was of the wrong
configuration (ivsize=4) for the task.
Anyway - it seems to me that it would be quite inappropriate to require
perl-5.6 support for a core perl module.

karenetheridge:
> Anyway - it seems to me that it would be quite inappropriate to require
perl-5.6 support for a core perl module.

List::Util is not core-only (only shipping with core), but rather dual-life (ships with core, but also installable separately via cpan). List::Util still supports perl 5.6 and ideally should continue to do so, whenever possible. Raising the minimum perl requirement on dual-life modules does happen, but generally only when unavoidable (for instance when unicode features are needed that were only added in 5.8 or 5.10, etc).

What can we do to resolve the failing tests on 5.6? (More people may be available to help after the long weekend is over.)

sisyphus:
On Mon, Sep 2, 2019 at 11:24 AM Karen Etheridge <notifications@github.com>
wrote:

> More people may be available to help after the long weekend is over

The more sets of eyes the better - and, of course, my proposed changes are
nothing more than a proposal.
It's fine by me if someone wants to come up with an alternative way of
addressing these failures in List::Util::uniqnum.

Looking at my Windows build of perl-5.6.2, I think I found the source of
the trouble.
While the value ~0 is assigned to a UV, for some reason perl-5.6.2 decides
to assign the value ~0 - 1 to an NV.

Assuming the same thing happens on a -Duse64bitint build of perl-5.6.2, the
value 18446744073709551615 - 1 is actually 1.8446744073709552e+19.
Of course, that's still a different value than 18446744073709551615 and the
failure of my patched build to detect that difference was a bug. (A "<="
needed to be a "<")

That bug has hopefully been fixed in the ListUtil.xs that I've just pushed.
I've also pushed another t/uniq.t that contains a test that would have
exposed that bug.

sisyphus:
On perl-5.30.0 (nvsize=8, ivsize=8), I've just noticed that List::Util::PP::uniqnum(~0, 2 ** 64) deems the second value to be a duplicate of the first.

Interestingly, perl agrees:
C:\> perl -le "print 'wtf' if ~0 == 2 ** 64;"
wtf

To my way of thinking, that is rubbish because (in binary) the first is '1' x 63 and the second is '1' . ('0' x 63).

Hence my patch to uniqnum  decrees that 2 ** 64 is *not* a duplicate of ~0.

Am I being unreasonable ?

sisyphus:
Duh ... '1' . ('0' x 63) should have been '1' . ('0' x 64).

sisyphus:
Heh ... and '1' x 63 should have been '1' x 64 ...

sisyphus:
Can someone explain to me what Travis CI is now complaining about ?

karenetheridge:
If you click on the details of each individual job that failed, it's fairly clear, e.g. at https://travis-ci.org/Dual-Life/Scalar-List-Utils/jobs/588407175:

```
curl: (6) Could not resolve host: cli-assets.heroku.com
xz: (stdin): File format not recognized
tar: Child returned status 1
tar: Error is not recoverable: exiting now
Failed to update Heroku CLI
```

That is, it's an issue upstream with the testing infrastructure.
You should be able to restart each failing job, or restart the entire build, from the travis interface.

sisyphus:
I had looked at the log, but couldn't see how it related to anything I had done. (Turns out it didn't, of course.)
All looks good again now - not that there was much wrong to begin with. It was working fine, but then I noticed it wasn't doing the job in the way that I intended.
Thanks Karen.

sisyphus:
Brief outline of the approach I've taken follows:

The success of uniqnum() XSub relies on its capacity to:
1) assign different strings to the SV "keysv" for different numeric values;
2) assign identical strings to the SV "keysv" for identical numeric values.
However, with the current Scalar-List-Utils-1.52, we sometimes see identical strings assigned for different numeric values, thereby contravening those conditions.
For example, uniqnum currently regards 1.4142135623730951 and 1.4142135623730954 as the same value, even though perl will correctly tell us that 1.4142135623730951 != 1.4142135623730954.

I've taken the approach of looking at the actual byte structure of the variables that hold the values, rather than the actual values themselves.
Clearly, if two NVs have the same byte structure then they encapsulate the same value.
Also, if two NVs have different byte structures then they encapsulate different values.

However, that last assertion is not always true - there are in fact some NVs that have different byte structures and yet encapsulate the same value.
Specifically, NaNs can have different byte structures; also the byte structure of the NV -0.0 differs from that of the NV 0.0.
I therefore handle the non-regular values (ie values that are either inf, nan, or zero) first - and set "keysv" to either "0", "NaN", "Inf" or "-Inf" as appropriate.
If the value is a regular value (i.e. neither NaN, nor an infinity nor zero) then "keysv" is set to the byte-structure (expressed as a hex string) of the NV representation of the value.
IOW, I do:

if(value of arg is an Inf, NaN, or zero) { 
  assign "Inf", "-Inf", "NaN" or "0" to "keysv", as appropriate;
}

else {
  assign the arg to an NV;
  assign the hex bytes of that NV to a string buffer;
  assign the string buffer to "keysv";
}

And that works fine for all perl configurations except the configuration where ivsize and nvsize are both 8 - the problem here being that this is the only perl configuration where IV/UV precision (64 bits) exceeds NV precision (53 bits).
For this (and only this) configuration, I define the symbol UNIQ_PREC_LOSS in ListUtil.xs and, in effect, proceed as follows:

if(value of arg is an Inf, NaN, or zero) { 
  assign "Inf", "-Inf", "NaN" or "0" to "keysv", as appropriate;
}

elsif(varg is a UV, and therefore greater than 2 ** 53) {
  assign the hex bytes of the UV to a string buffer;
  cast the UV to an NV, and concatenate the hex bytes of that NV onto the string buffer;
  assign the string buffer to "keysv";
}

elsif(varg is an IV && (value > 2 ** 53 || value < -(2 ** 53)) ) {
  assign the hex bytes of the IV to a string buffer;
  cast the IV to an NV, and concatenate the hex bytes of that NV onto the string buffer;
  assign the string buffer to "keysv";
}

elsif(arg expressed as an NV is an integer value && (
                                    (value < 2 ** 64 && value > 2 ** 53)
                                      ||
                                    (value < -(2 ** 53) && value  >= -(2 ** 63))
                                                     )
                                                       ) {
  assign the hex bytes of the IV/UV representation of the value to the string buffer;
  concatenate the hex bytes of the NV representation of the value onto the string buffer;
  assign the string buffer to "keysv";
}

else {
  assign the arg to an NV;
  assign the hex bytes of that NV to a string buffer;
  assign the string buffer to "keysv";
}

Note that this procedure differs from the first only in the inclusion of the three "elsif" blocks.
At one stage, for the case where ivsize and nvsize are both 8, I considered simply assigning to "keysv" the IV/UV hex bytes concatenated with NV hex bytes for *all* regular values, irrespective of data type and magnitude.
However, (eg) coercing a double that's greater than or equal to 2 ** 64 to a UV is undefined behavior in C99, and ought therefore be avoided.

Cheers,
Rob

9007199254740992       == 2 ** 53
9.2233720368547758e+18 == 2 ** 63
1.8446744073709552e+19 == 2 ** 64

leonerd:
I'm unconvinced at the moment. This seems to be quite a large change to fix what I'm still not quite convinced is actually a bug.

Yes, it is true that `uniqnum()` can consider two values equal when `==` does not, or vice versa; but that only happens down in the "weeds" of very similar floating-point values. (Though I will accept if it handles 0.0 vs -0.0 differently that certainly is a bug).

As you've seen in trying to implement this, the actual function's behaviour can differ in really subtle circumstances determined by perl compiletime options visible in the `%Config` hash. As a result, you've created a function that may be hard to reason about.

Is there an actual practical reason why the behaviour needs fixing - some actual bug in a real program which this change resolves? If not, I wonder if a far shorter, simpler solution would be simply to document that as it involves floating-point values, `uniqnum` may not always exactly agree with the `==` operator on very simliar values in the lowest significant digits. I feel that for any practical purpose, if this really did matter to someone's logic, they wouldn't be using floats in the first place but instead something like an arbitrary-precision `Math::BigRat`, and thus would be using stringwise uniqueness testing and their code would behave more sensibly.

However, even if we take at face-value the necessity of making this change, I have a few concerns about the implementation of it. First off - the logic seems to apply in all value cases, not just the floating-point case. This feels like it will have a performance impact even when operating on lists of pure integers. Can you convince me this will be OK?

Secondly, your exact method of generating a uniqueness key for floats appears to be to generate a hex string encoding the raw underlying implementation bytes. I don't see any safety logic preventing a floating-point value whose hex string happens to contain only the digits 0 to 9 from colliding with an integer whose decimal expansion consists of exactly those same digits. Suddenly it is now possible to construct a pair of a floating-point and large-integer value which perl considers to be (very) different values, yet this code generates the same collision key and thus one is rejected from the uniqueness filter.

karenetheridge:
> If not, I wonder if a far shorter, simpler solution would be simply to document that as it involves floating-point values, uniqnum may not always exactly agree with the == operator on very simliar values in the lowest significant digits.

I agree.  Comparing two floating point numbers for exactness has always been a bad idea in non-base-10 architectures (i.e. all of them). The only reasonable comparison to make is (paraphrased) `uniq_by { round_to_n_decimal_places($_) }`.

sisyphus:
On Tue, Oct 8, 2019 at 2:29 AM Paul Evans <notifications@github.com> wrote:

> Yes, it is true that uniqnum() can consider two values equal when == does
> not, or vice versa; but that only happens down in the "weeds" of very
> similar floating-point values. (Though I will accept if it handles 0.0 vs
> -0.0 differently that certainly is a bug).
>
To me, the notion that 2 floats that are arbitrarily close in value should
be considered duplicates is both insane, unwise and unnecessary.
Users that wish to cater for such a condition should write their own code
to handle such wishes, and they should have the right to decide for
themselves the size of the discrepancy below which unequal floats should be
considered equivalent.
As it stands, version 1.52 is imposing the condition any two NVs (be they
double, long double or __float128) that round to the same 15-decimal-digit
string are to be considered equivalent.
I think that takes a bad idea to an even worse extreme.

As it turns out, this problem can be avoided (in the case of 8-byte NVs)
by, in ListUtil.xs, changing:
sv_setpvf(keysv, "%.15" NVgf, SvNV(arg));
to
sv_setpvf(keysv, "%.17" NVgf, SvNV(arg));

That change enables unequal doubles to be correctly identified as unequal.
(It doesn't fix the problem with -0.0, but that can be dealt with
separately.)

The problem we're then left with is that, if ivsize is 8 (as is mostly the
case), there's a large number of integer value pairs that are equivalent,
but are identified in 1.52 by uniqnum() as being unequal. (This is still a
problem even if we leave the required digits at 15.)
For example, uniqnum regards 9223372036854775808 and 9223372036854775808.0
as unequal - though both are exact representations of 2 ** 63; the former
being an IV, the latter an NV.
You can keep on dividing both values by 2  repeatedly, all the way down to
1125899906842624 and 1125899906842624.0 (2 ** 50) and uniqnum will still
say they're unequal.
And you can add some of those values together and uniqnum will still state
that they are unequal.
When "arg" contains a small integer value, ["%" IVdf, SvIV(arg)] and ["
%.17" NVgf, SvNV(arg)] provide identical strings which enables uniqnum to
work correctly when comparing IVs and NVs but, for larger values (such as
those just given), the 2 different format arguments start providing
different strings (eg "9223372036854775808" and "9.22337203685478e+18").
And this breaks uniqnum.

This is what led me to start looking at both IV and NV byte structures -
but, I think you're right in that there's a better way.
It's probably obvious to anyone with a brain, but it took me until today to
observe that this second issue can be taken care of, in ListUtil.xs by
setting the number of digits required to 19:

sv_setpvf(keysv, "%.19" NVgf, SvNV(arg));

With that change in place, all issues (except the one regarding -0.0) are
taken care of for the case where nvsize is 8.
And the version of uniq.t provided by my pull request, passes all tests
(except test 25, which tests handling of -0.0).

With extended precision long doubles, I'm getting passes (for all but the
-0.0 test) with:
sv_setpvf(keysv, "%.21" NVgf, SvNV(arg));

And for the NV type of __float128 (and IEEE long double), we need:

sv_setpvf(keysv, "%36" NVgf, SvNV(arg));

This also tests fine for me.

For the double-double, using values is quite problematic, so I think I'll
stick with looking at the byte structure there.

All of this should be readily achievable via preprocessing, but it could
take me a couple of days to get it set up and tested, as I have a bit on
tomorrow.


>  I don't see any safety logic preventing a floating-point value whose hex
> string happens to contain only the digits 0 to 9 from colliding with an
> integer whose decimal expansion consists of exactly those same digits.


That has the ring of being a valid complaint. I don't have an example of
such a case but, sadly, I suspect I could come up with one if I set my mind
to it.

I'll get back in a day or 2 with (hopefully) something that's more
acceptable.

Thanks for the help.

Cheers,
Rob

sisyphus:
Given that I'm about to substantially rewrite the pull request, I suspect it would be more helpful (wrt readability and digestibility) if the next patches generated for ListUtil.xs and Makefile.PL are written against the versions of those files that are in the current github repo (instead of written against the versions of those files that currently appear in my fork).

Am I right in my thinking on this ?
If so, what's the best way to achieve such an outcome ?

Cheers,
Rob

sisyphus:
Given that my next attempt is quite different to the approach taken here, I'm closing this pull request.
I'll then delete my current fork of sisyphus/Scalar-List-Utils, and start over afresh, with a new fork and a new pull request.
Hope this all works as intended ....
Cheers,
Rob

PR #80, opened 2019-10-12
Fix uniqnum

sisyphus:
Another attempt to overcome current inadequacies in List::Util::uniqnum().
At this stage I'm just wanting to check that travis doesn't find anything objectionable.

That being so, there's still an issue in getting uniqnum() to work correctly on Windows for perls earlier than 5.26.0. (It's not a new issue, btw, as it already exists on those perls, even if current 1.52 is installed.)

There's also an unclear issue (again not new) with Debian and Ubuntu - for which I've filed a perlbug report, but for which no ticket has yet been assigned.

Both of those caveats exist around getting, for example, uniqnum(1 << 59, 2 ** 59) to recognize that both values are the same.
AFAICT, uniqnum has never, since its inception, recognized this.

Cheers,
Rob

### DIFF
--- a/ListUtil.xs
+++ b/ListUtil.xs
@@ -85,6 +85,21 @@
 #error "LU_NV_BITS not set"
 #endif
 
+#if defined(FALLBACK_TO_BYTES) || defined(NV_IS_DOUBLEDOUBLE)
+
+#define IF_NAN_ZEROS_INFS \
+            if(nv_arg != nv_arg) {				\
+                sv_setpvf(keysv, "%s", "NaN");			\
+            }							\
+            else if(nv_arg == 0) {				\
+                sv_setpvf(keysv, "%s", "0");			\
+            }							\
+            else if(nv_arg / nv_arg!=1) {			\
+                if(nv_arg < 0)sv_setpvf(keysv, "%s", "-Inf");	\
+                else sv_setpvf(keysv, "%s", "Inf");		\
+            }							
+#endif
+
 /* Some platforms have strict exports. And before 5.7.3 cxinc (or Perl_cxinc)
    was not exported. Therefore platforms like win32, VMS etc have problems
    so we redefine it here -- GMB
@@ -1177,7 +1192,7 @@ CODE:
     int i;
 #endif
 #ifdef FALLBACK_TO_BYTES /* Defined by Makefile.PL */
-    size_t potential_prec_loss = 0, offset = 0;
+    size_t potential_prec_loss, offset;
     IV int_arg;
 #endif
 
@@ -1216,22 +1231,9 @@ CODE:
 
             nv_arg = SvNV(arg);
 
-            /* Handle NaN, zeros and Infs */
+            IF_NAN_ZEROS_INFS /* Handle Nan, Zero or Inf values */
 
-            if(nv_arg != nv_arg) {
-                sv_setpvf(keysv, "%s", "NaN");
-            }
-
-            else if(nv_arg == 0) {
-                sv_setpvf(keysv, "%s", "0");
-            }
-
-            else if(nv_arg/nv_arg != 1) {
-                if(nv_arg < 0) sv_setpvf(keysv, "%s", "-Inf");
-                else sv_setpvf(keysv, "%s", "Inf");
-            }
-
-            else {
+            else {            /* Handle values other NaN, Zeros, Infs */
                 for(i = 0; i < 16; i++) {
                     sprintf(buffer, "%02x", ((unsigned char*)p)[i]);
                     buffer += 2;
@@ -1240,8 +1242,11 @@ CODE:
                 sv_setpvf(keysv, "%s", buffer);
             }
 #elif defined(FALLBACK_TO_BYTES) /* Defined by Makefile.PL */
+            potential_prec_loss = 0; /* clear */
+            offset = 0;              /* clear */
+            nv_arg = SvNV(arg);
+            
             if(!SvOK(arg) || SvUOK(arg)) {
-                nv_arg = SvNV(arg);
                 int_arg = SvIV(arg); /* SvIV(arg) and SvuV(arg) have the same byte structure *
                                       * and it's only the byte structure that interests us   */
 
@@ -1249,7 +1254,6 @@ CODE:
             }
 
             else if(SvIOK(arg)) {
-                nv_arg = SvNV(arg);
                 int_arg = SvIV(arg);
 
                 if(int_arg < -9007199254740992 
@@ -1258,30 +1262,9 @@ CODE:
                     potential_prec_loss = 1;
             }
 
-            else {
-                nv_arg = SvNV(arg);
-            }
+            IF_NAN_ZEROS_INFS /* Handle NaN, Zero or Inf values */
 
-            /* Handle NaN, zeros and Infs */
-            if(nv_arg != nv_arg) {
-                sv_setpvf(keysv, "%s", "NaN");
-                potential_prec_loss = 0; /* clear */
-            }
-
-            else if(nv_arg == 0) {
-                sv_setpvf(keysv, "%s", "0");
-                potential_prec_loss = 0; /* clear */
-            }
-
-            else if(nv_arg/nv_arg != 1) {
-                if(nv_arg < 0) sv_setpvf(keysv, "%s", "-Inf");
-                else sv_setpvf(keysv, "%s", "Inf");
-                potential_prec_loss = 0; /* clear */
-            }
-
-            /* Handle all values other than NaN, zeros and Infs */
-
-            else {
+            else {            /* Handle values other NaN, Zeros, Infs */
 
                 if(potential_prec_loss) { 
 
@@ -1308,8 +1291,6 @@ CODE:
                     offset = 16; /* the buffer pointer has been incremented by 16 */
                 }
 
-                potential_prec_loss = 0; /* clear */
-
                 /* Read the bytes of SvNV(arg) into buffer, in hex.    *
                  * If buffer already contains the bytes of SvIV(arg)   *
                  * then this action appends the hex bytes of SvNV(arg) *
@@ -1321,8 +1302,6 @@ CODE:
                 }
 
                 buffer -= 16 + offset; /* return pointer to original position */
-                offset = 0;
-
                 sv_setpvf(keysv, "%s", buffer);
             }
 #else
--- a/ListUtil.xs
+++ b/ListUtil.xs
@@ -85,19 +85,12 @@
 #error "LU_NV_BITS not set"
 #endif
 
-#if defined(FALLBACK_TO_BYTES) || defined(NV_IS_DOUBLEDOUBLE)
-
-#define IF_NAN_ZEROS_INFS \
-            if(nv_arg != nv_arg) {				\
-                sv_setpvf(keysv, "%s", "NaN");			\
-            }							\
-            else if(nv_arg == 0) {				\
-                sv_setpvf(keysv, "%s", "0");			\
-            }							\
-            else if(nv_arg / nv_arg!=1) {			\
-                if(nv_arg < 0)sv_setpvf(keysv, "%s", "-Inf");	\
-                else sv_setpvf(keysv, "%s", "Inf");		\
-            }							
+#if defined(NV_IS_DOUBLEDOUBLE)
+#define UNIQ_LOOP 16
+#define UNIQ_BUFFER_RESET 32
+#elif defined(FALLBACK_TO_BYTES)
+#define UNIQ_LOOP 8
+#define UNIQ_BUFFER_RESET 16 + offset
 #endif
 
 /* Some platforms have strict exports. And before 5.7.3 cxinc (or Perl_cxinc)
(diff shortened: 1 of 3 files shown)


sisyphus:
BTW, I've decided that there was really nothing wrong with the fix that Issue 79 provided - save that it can be up to about 30% slower than the error-prone implementation that is in 1.52.
It was suggested in Issue 79 that, when comparing an IV to an NV by referring to their byte structures, we might incorrectly decide that an IV and an NV that have identical byte structures are the same value, even though they're not.
I conceded that might be possible - but in so doing, I was making a mistake.
The thing is that, whenever I took the byte structure of SvIV(arg), I always appended the byte structure of SvNV(arg) onto it.
The SV keysv will never get set to the hex-byte structure of an IV/UV - though there are situations when keysv will be set to the concatenation of the hex-byte structure of the IV/UV representation with the hex-byte structure of the NV representation.

This becomes relevant with my latest changes here in Issue 80, where I've had to resort to the very same process of examining bytes - but this time only for Windows builds where ivsize == 8 && nvsize == 8 && ($] < 5.026 || perl was not built with -D__USE_MINGW_ANSI_STDIO).
This is the case where FALLBACK_TO_BYTES is defined (by the Makefile.PL).
To make things more difficult, there's no %Config key that we can query to determine whether perl was built with -D__USE_MINGW_ANSI_STDIO - though I know that Strawberry Perl is built *with* it.

In the first version of ListUtil.xs that appeared in this Issue 80 pull request, there was a section where I had:

sprintf(buff, "%.19g", SvNV(arg));
sv_setpvf(keysv, "%s", buff);

even though, for many perl configurations, the more succinct rendition of:

sv_setpvf(keysv, "%.19g", SvNV(arg);

would have worked just as well.
In this latest version of ListUtil.xs the 2 line version is pre-processed away (by defining GCVT_WIDTH_LIMIT) unless the 2 line version is really needed.
I've based this on information provided to me by Haarg at:
https://www.perlmonks.org/?node_id=11107370
If this is a case of splitting a hair that doesn't really need to be split, I'm quite happy to do away with the one line rendition altogether, and just settle for the more generic 2 line rendition.

Is there anything else I need to do ?

As regards the notion that NV values that are close to the same value ought to be regarded as duplicates, consider the following:
Take, for example, the 2 doubles represented by 0x1.6a09e667f3bcep0 and 0x1.6a09e667f3bcdp0.
The former is accurately, exactly, precisely equal to the rational 6369051672525774/4503599627370496 and the latter is accurately, exactly, precisely equal to the rational 6369051672525773/4503599627370496.
The values are indeed close to each other (for some arbitrary measure of closeness), but to assert that both doubles are equivalent is also to assert that both rationals are equivalent.
I really cannot see any reason that such assertions should be made.

Cheers,
Rob

sisyphus:
Haarg has just found a way that I can dispense with GCVT_WIDTH_LIMIT and reduce clutter at zero cost.
See https://www.perlmonks.org/?node_id=11107410

I'll push that fix tomorrow.

sisyphus:
In addition to the removal of clutter (as mentioned in my last post) I built perl-5.31.4 on Windows without -D__USE_MINGW_ANSI_STDIO just so I could check that I was detecting such builds in the Makefile.PL.
Turns out I wasn't. I needed to change the condition

sprintf("%.0f", 2 ** 63) =~ /0$/
to
sprintf("%.0f", 2 ** 64) =~ /0$/

So I've done that, too.

I don't see anything else that needs attention.
Please point out anything I've missed.

It might be preferable to wait until 5.31.5 has been released (only 6 days away) before updating Scalar-List-Utils.
Then I get plenty of time to deal with any other issues that the smokers throw up before the next blead release.

I've tested these Scalar-List-Utils changes on a number of systems and perl configurations.

On Debian Wheezy, I've tested:
perl-5.31.4, nvtype = double
perl-5.31.4, nvtype = long double (DoubleDouble)

On Ubuntu-18.04, I've tested:
perl-5.31.4, nvtype = double
perl-5.31.4, nvtype = long double
perl-5.31.4, nvtype = __float128

On freebsd-12.0, I've tested:
perl-5.31.4, nvtype = double
perl-5.31.4, nvtype = long double
32-bit build of perl-5.30.0, nvtype = long double, ivsize = 4

On Windows 7 I've tested:
32-bit build of perl-5.8.8, nvtype = double, ivsize = 4
32-bit build of perl-5.10.0, nvtype = double, ivsize = 4
32-bit build of perl-5.12.0, nvtype = double, ivsize = 4
64-bit build of perl-5.12.0, nvtype = double, ivsize = 4
32-bit build of perl-5.14.0, nvtype = double, ivsize = 4
64-bit build of perl-5.14.0, nvtype = double, ivsize = 4
32-bit and 64-bit builds of perl-5.16.0, nvtype = double
32-bit and 64-bit builds of perl-5.18.0, nvtype = double
32-bit and 64-bit builds of perl-5.20.0, nvtype = double
32-bit and 64-bit builds of perl-5.22.0, nvtype = double
32-bit and 64-bit builds of perl-5.22.0, nvtype = long double
32-bit and 64-bit builds of perl-5.24.0, nvtype = double
32-bit and 64-bit builds of perl-5.24.0, nvtype = long double
32-bit and 64-bit builds of perl-5.26.0, nvtype = double
32-bit and 64-bit builds of perl-5.26.0, nvtype = long double
32-bit and 64-bit builds of perl-5.28.0, nvtype = double
32-bit and 64-bit builds of perl-5.28.0, nvtype = long double
32-bit and 64-bit builds of perl-5.30.0, nvtype = double
32-bit and 64-bit builds of perl-5.30.0, nvtype = long double
32-bit and 64-bit builds of perl-5.31.4, nvtype = double
32-bit and 64-bit builds of perl-5.31.4, nvtype = double, non-ansi
32-bit and 64-bit builds of perl-5.31.4, nvtype = long double
Also, a few builds of Strawberry Perl from over the last 4 years.

On all builds, ivsize was 8 unless otherwise specified.
The Windows builds are all built using mingw toolchain (gcc compiler).
It's rare, these days, that perl gets built on Windows using a  Microsoft Compiler.
I know that Steve Hay still takes an active interest in MSVC builds of perl, and I'll ping him when these changes hit bleadperl to ensure some testing on such builds, as I don't think they are being smoked.

Any other queries, concerns - just ask.

Cheers,
Rob

haarg:
This patch seems like it has a lot of duplication in it that shouldn't be needed.

sisyphus:
I don't see any duplication that's impacting on performance. (Please let me
know if you've spotted some.)
But I can rearrange a couple of bits and pieces to significantly reduce the
amount of written code - which I'll embark upon now.

sisyphus:
There's still a couple of sections of code that are near-duplicates.

In the NV_IS_DOUBLEDOUBLE block we have:

###########################
for(i = 0; i < 16; i++) {
  sprintf(buffer, "%02x", ((unsigned char*)p)[i]);
  buffer += 2;
}
buffer -= 32;
sv_setpvf(keysv, "%s", buffer);
###########################

and in the FALLBACK_TO_BYTES block we have:

###########################
for(i = 0; i < 8; i++) {
  sprintf(buffer, "%02x", ((unsigned char*)p)[i]);
  buffer += 2;
}

buffer -= 16 + offset; /* return pointer to original position */
sv_setpvf(keysv, "%s", buffer);
###########################

They could both be replaced with a macro named (eg) BYTES_TO_KEYSV (untested):

###########################
#if defined(NV_IS_DOUBLEDOUBLE)
#define UNIQ_LOOP 16
#define BUFFER_RESET 32
#elif defined(FALLBACK_TO_BYTES)
#define UNIQ_LOOP 8
#define BUFFER_RESET 16 + offset
#endif

#if defined(NV_IS_DOUBLEDOUBLE) || defined(FALLBACK_TO_BYTES)
#define BYTES_TO_KEYSV                             \
for(i = 0; i < LOOP; i++) {                        \
  sprintf(buffer, "%02x", ((unsigned char*)p)[i]); \
  buffer += 2;                                     \
}                                                  \
buffer -= BUFFER_RESET;                            \
sv_setpvf(keysv, "%s", buffer); 
#endif
###########################

(Seems that the line-continuation backslashes for the BYTES_TO_KEYSV macro are removed by the github rendering of this post.) 

If I was trying to read through the code for the first time, I'd probably find the presence of that BYTES_TO_KEYSV macro to be quite annoying.
In fact, I wonder if the same complaint could be leveled at the IF_NAN_ZEROS_INFS macro that I did implement.

If it's deemed preferable to include the BYTES_TO_KEYSV macro (or, indeed, to remove the IF_NAN_ZEROS_INFS macro), just let me know and I'll oblige without question or complaint.

Or maybe I'm still missing something ?

Cheers,
Rob

sisyphus:
Ok - that's it for me.
The latest changes to ListUtil.xs were:
1) Some re-arrangement that has no effect on functionality;
2) Remove an errant comment, and an unnecessary assignment to the IV int_arg.
3) Remove an unnecessary check that determines if arg is an Inf.

Case 2) relates only to FALLBACK_TO_BYTES being defined - ie where the OS is MS Windows && "%.19g" fails to deliver the requisite accuracy by coercing any digits after the 17th to 0.
I don't discount the possibility that other systems might exist for which FALLBACK_TO_BYTES needs to be defined - having tested only Debian Wheezy, Ubuntu-18.04, freebsd-12.0 and Windows 7.

Case 3) relates only to NV_IS_DOUBLEDOUBLE or FALLBACK_TO_BYTES being defined.
The check was originally there to guard against the possibility of doing SvIV(arg) when arg is an NV whose value is an infinity. I believe that would be undefined behaviour in C, but the current code won't ever call SvIV(arg) if arg is an infinity.

In Makefile.PL I made only a couple of changes to the opening comments that I've provided near the start of that file.

Other than that, feel free to modify or replace elements of this pull request as you see fit.
I'm not at all precious about that.
All I want to see is that Scalar-List-Utils pass the modified version of t/uniq.t that I've provided as part of this pull request.

Cheers,
Rob

haarg:
What platforms and perl configurations are you testing this on?  I'd like to test this, and possible some modifications of it, but I want to be sure I'm covering all of the relevant platforms.

sisyphus:
@haarg,
I'm hopeful that this approach would work on all platforms - for recent perls at least.

As indicated by https://www.nntp.perl.org/group/perl.perl5.porters/2019/11/msg256605.html , there's a problem with perl-5.14 on Tux's HP-UX-11.11 system in the way that it stringifies inf and/or nan.
The code I've supplied won't fix that. It's essentially a bug in the test script that could be handled in a number of ways - eg by skipping the failing test (as I do for a similar problem with old perls on Windows), or by altering the test to avoid Nan and Inf values.

I've listed above (in my Oct 15 comment) many of the configurations and perl versions that I've tested.
Not sure what you mean by "relevant platforms" - I think all platforms are relevant.

Please do take a look, and please feel free to make alterations.

Cheers,
Rob

haarg:
Sorry, I missed the list of configurations you had posted above.  By 'relevant' I meant what platforms I would need to check to cover the different the different NV formats this PR is meant to handle.

sisyphus:
The NV formats that I'm aware of are 8-byte double, 8-byte long double, extended precision (10-byte) long double, IEEE (16-byte) long double,  IBM (16-byte DoubleDouble) long double, 16-byte __float128.

I believe the PR covers all of them, and for all endianesses, though I've not specifically tested all permutations.
For example, I don't have access to an architecture that provides the 16-byte IEEE long double, but I believe such architectures are covered because my testing on -Dusequadmath builds (where vnvtype is __float128) reveals no issues.

In a nutshell, I can't think of any configuration that could pose a problem, except for those that stringify inf and/or nan  bizarrely,
Even then, the problem is one of how to properly construct tests. 

I think, just test what you've got access to ;-)

Cheers,
Rob

haarg:
@sisyphus Can you test this branch of List::Util::PP on the perls you have available that are not nvsize=8 and ivsize=8? https://github.com/haarg/List-Util-MaybeXS/tree/new-uniqnum

sisyphus:
@haarg,
Ummm ... not so sure this is really the place to be looking at List-Util-MaybeXS.
Are you confusing this thread with https://github.com/haarg/List-Util-MaybeXS/pull/2 ?

Anyway, the new-uniqnum branch of List-Util-MaybeXS tests fine.
I tested the cases:
ivsize = 4, nvmantbits = 52
ivsize = 4, nvmantbits = 64
ivsize = 8, nvmantbits = 64
ivsize = 8, nvmantbits = 112 (__float128)
ivsize = 8, nvmantbits =  104 (DoubleDouble)

Cheers,
Rob

sisyphus:
I would find it ever so nice if someone with a Scalar-List-Utils commit bit could take some action on this.
Options (in descending order of my preference) are as follows:

1) Apply my pull request as is;
2) Apply a modified version of my pull request such that the tests provided by my pull-request's version of t/uniq.t still pass;
3) Give me a commit bit so that I can apply my pull request;
4) Modify the existing version of uniqnum such that inclusion of NVs is a fatal error, and modify uniqnum documentation & t/uniq.t accordingly;
5) Remove uniqnum from the Scalar-List-Utils source distro;
6) Remove Scalar-List-Utils from perl core.

I believe that the enaction of either one of those options would be sufficient to shut me up.

Option 4) arises because, AFAIK, there are no problems with uniqnum when it involves only IVs and/or UVs.
That is, absurd results can arise only when NVs are involved.

A 7th way of shutting me up would be to prove to my satisfaction that my pull-request's version of t/uniq.t is unreasonable in its expectations.
I didn't put that in the list because I believe it's impossible to convince me of that.

Cheers,
Rob

leonerd:
I believe #83 takes option 2. It includes the tests here but is implemented in a neater fashion which is easier to follow.

sisyphus:
Thanks - I was unaware of the existence of that PR.
I'll post my follow ups over on #83.

PR #82, opened 2019-10-30
isstring and isnumeric?

jo-37:
The two subs "isstring" and "isnumeric" would nicely fit into Scalar::Util IMHO.
What do you think about it?


leonerd:
I'm not sure they are useful *yet*, while the P/I/NOK flags are quite so fragile in core perl. But hopefully soon when these are a little less fragile, and can stand to express programmer intent as to what the primary nature of a value is, they will become useful

leonerd:
I'm going to close this one now, as in the three years since the first thoughts many things have happened. Most notably, we now have the `builtin::` core namespace where things like this should live.

There's an [RFC](https://github.com/Perl/RFCs/pull/13) and a core perl [PR](https://github.com/Perl/perl5/pull/19524) to add some functions similar to this, however.

PR #83, opened 2019-11-30
increase accuracy of uniqnum

haarg:
Using sprintf "%.15g" as the value to compare does not maintain the full
accuracy of floating point numbers.  Rather than 15, use a precision
value based on the accuracy the current platform can represent.

Also add some special cases: treat 0 and -0.0 as equivalent, and NaNs
are compared based on the platform's stringification of them.

Additionally, sprintf may not be able to accurately represent some
floating point numbers, no matter how many digits are given.  If the
numbers are outside the IV/UV range, they don't need to be compared to
anything other than other floats, so we can use a totally different
format for the unique string.  Use their raw bytes, as this is the
easiest and fastest.

Fixes RT#130935, RT#130302, RT#130277

### DIFF
--- a/ListUtil.xs
+++ b/ListUtil.xs
@@ -72,9 +72,6 @@
 #define sv_catpvn_flags(b,n,l,f) sv_catpvn(b,n,l)
 #endif
 
-#define STRINGIZE_HELPER(x) #x
-#define STRINGIZE(x) STRINGIZE_HELPER(x)
-
 /* Some platforms have strict exports. And before 5.7.3 cxinc (or Perl_cxinc)
    was not exported. Therefore platforms like win32, VMS etc have problems
    so we redefine it here -- GMB
@@ -1245,7 +1242,7 @@ CODE:
                 /* smaller floats get formatted using %g and could be equal to
                  * a UV or IV */
                 else {
-                    sv_setpvf(keysv, "%0." STRINGIZE(NV_MAX_PRECISION) NVgf, nv_arg);
+                    sv_setpvf(keysv, "%0.*" NVgf, NV_MAX_PRECISION, nv_arg);
                 }
             }
 #ifdef HV_FETCH_EMPTY_HE
--- a/ListUtil.xs
+++ b/ListUtil.xs
@@ -1235,9 +1235,11 @@ CODE:
                     sv_setpvf(keysv, "%" NVgf, nv_arg);
                 }
                 /* for numbers outside of the IV or UV range, we don't need to
-                 * use a comparable format, so just use the raw bytes */
+                 * use a comparable format, so just use the raw bytes, adding
+                 * 'f' to ensure not matching a stringified number */
                 else if (nv_arg < (NV)IV_MIN || nv_arg > (NV)UV_MAX) {
                     sv_setpvn(keysv, (char *) &nv_arg, sizeof(NV));
+                    sv_catpvn(keysv, "f", 1);
                 }
                 /* smaller floats get formatted using %g and could be equal to
                  * a UV or IV */
--- a/ListUtil.xs
+++ b/ListUtil.xs
@@ -72,6 +72,9 @@
 #define sv_catpvn_flags(b,n,l,f) sv_catpvn(b,n,l)
 #endif
 
+#define STRINGIZE_HELPER(x) #x
+#define STRINGIZE(x) STRINGIZE_HELPER(x)
+
 /* Some platforms have strict exports. And before 5.7.3 cxinc (or Perl_cxinc)
    was not exported. Therefore platforms like win32, VMS etc have problems
    so we redefine it here -- GMB
@@ -1201,6 +1204,7 @@ CODE:
 
         for(index = 0 ; index < items ; index++) {
             SV *arg = args[index];
+            NV nv_arg;
 #ifdef HV_FETCH_EMPTY_HE
             HE* he;
 #endif
@@ -1217,12 +1221,33 @@ CODE:
 #endif
             }
 
-            if(!SvOK(arg) || SvUOK(arg))
+            if(!SvOK(arg) || SvUOK(arg)) {
                 sv_setpvf(keysv, "%" UVuf, SvUV(arg));
-            else if(SvIOK(arg))
+            }
+            else if(SvIOK(arg)) {
                 sv_setpvf(keysv, "%" IVdf, SvIV(arg));
-            else
-                sv_setpvf(keysv, "%.15" NVgf, SvNV(arg));
+            }
+            else {
+                nv_arg = SvNV(arg);
+                /* use 0 for both 0 and -0.0 */
+                if(nv_arg == 0) {
+                    sv_setpvs(keysv, "0");
+                }
+                /* for NaN, use the platform's normal stringification */
+                else if (nv_arg != nv_arg) {
+                    sv_setpvf(keysv, "%" NVgf, nv_arg);
+                }
+                /* for numbers outside of the IV or UV range, we don't need to
+                 * use a comparable format, so just use the raw bytes */
+                else if (nv_arg < (NV)IV_MIN || nv_arg > (NV)UV_MAX) {
+                    sv_setpvn(keysv, (char *) &nv_arg, sizeof(NV));
+                }
+                /* smaller floats get formatted using %g and could be equal to
+                 * a UV or IV */
+                else {
+                    sv_setpvf(keysv, "%0." STRINGIZE(NV_MAX_PRECISION) NVgf, nv_arg);
+                }
+            }
 #ifdef HV_FETCH_EMPTY_HE
             he = (HE*) hv_common(seen, NULL, SvPVX(keysv), SvCUR(keysv), 0, HV_FETCH_LVALUE | HV_FETCH_EMPTY_HE, NULL, 0);
             if (HeVAL(he))
--- a/Makefile.PL
+++ b/Makefile.PL
@@ -6,12 +6,42 @@ use Config;
 use File::Spec;
 use ExtUtils::MakeMaker;
 my $PERL_CORE = grep { $_ eq 'PERL_CORE=1' } @ARGV;
+my $defines = $ENV{PERL_CORE} ? q[-DPERL_EXT] : q[-DPERL_EXT -DUSE_PPPORT_H];
+
+my $nv_digits;
+
+my $ivsize = $Config{ivsize} * 8;
+my $uv_digits = 1 + int(log(2) / log(10) * $ivsize);
+
+my $nvsize = $Config{nvsize} * 8;
+# we want NV and UV numbers in the same range to format the same, so make sure
+# NVs are given at least as many digits as UVs.  If IV/UVs have equal or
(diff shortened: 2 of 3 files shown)


haarg:
This is an alternative to #79 and #80 that I believe should fix the same issues, but is significantly easier to understand.

ListUtil.xs
```diff
@@ -1217,12 +1221,33 @@ CODE:
...
+                }
+                /* for numbers outside of the IV or UV range, we don't need to
+                 * use a comparable format, so just use the raw bytes */
+                else if (nv_arg < (NV)IV_MIN || nv_arg > (NV)UV_MAX) {
+                    sv_setpvn(keysv, (char *) &nv_arg, sizeof(NV));
+                }
+                /* smaller floats get formatted using %g and could be equal to
+                 * a UV or IV */
+                else {
+                    sv_setpvf(keysv, "%0." STRINGIZE(NV_MAX_PRECISION) NVgf, nv_arg);
```
leonerd: Rather than this `STRINGIZE` macro, you can just use the "%0.*g" ability of `sprintf`:

```
    sv_setpvf(keysv, "%0.*" NVgf, NV_MAX_PRECISION, nv_arg);
```

Changing the code to this keeps all the tests passing for me; you can then delete the macro at the top.

ListUtil.xs
```diff
@@ -1217,12 +1221,33 @@ CODE:
...
+            }
+            else {
+                nv_arg = SvNV(arg);
+                /* use 0 for both 0 and -0.0 */
+                if(nv_arg == 0) {
+                    sv_setpvs(keysv, "0");
+                }
+                /* for NaN, use the platform's normal stringification */
+                else if (nv_arg != nv_arg) {
+                    sv_setpvf(keysv, "%" NVgf, nv_arg);
```
leonerd: I don't quite like this as is, because it remains possible (however unlikely) to construct a floating-point number whose bytes happen to be all ASCII digits:

```
eval: pack "F", 6.82132005170133e-38
'62345678'
```

and thus they'd compare equal using this algorithm. Can you prepend or suffix a non-digit, perhaps "r" for "raw", in order to avoid this?

leonerd:
LGTM. Thanks.

sisyphus:
On Windows, when ivsize and nvsize are both 8, and perl was not built with -D__USE_MINGW_ANSI_STDIO we get:

not ok 19 - uniqnum correctly compares UV/IVs that don't overflow NVs
\#   Failed test 'uniqnum correctly compares UV/IVs that don't overflow NVs'
\#   at t/uniq.t line 190.
\#     Structures begin differing at:
\#          $got->[1] = '9.22337203685478e+018'
\#     $expected->[1] = '1152921504606846976'

Prior to 5.26 it was not possible to build perl on Windows with -D__USE_MINGW_ANSI_STDIO, so that means that 5.24 and earlier are affected (but only when ivsize and nvsize are both 8).

The problem is related to the way that these perls stringify large integer doubles - best exemplified by the following.
On perl-5.30.0 (built with -D__USE_MINGW_ANSID_STDIO):

C:\>perl -le "printf '%.0f', 2 ** 64;"
18446744073709551616

On perls built without -D__USE_MINGW_ANSI_STDIO:
C:\>perl -le "printf '%.0f', 2 ** 64;"
18446744073709552000

It's not just powers of 2 that are affected.

The only way I could think of dealing with this was to instead look at the bytes (both as an NV & IV), and that's why FALLBACK_TO_BYTES existed in the version I came  up with.

There's no way to query Config to find out whether or not __USE_MINGW_ANSI_STDIO was defined.
The test I applied (only to Windows perls whose ivsize and nvsize were 8) was:

if($] < 5.026 || sprintf("%.0f", 2 ** 64) =~ /0$/) {
  $defines .= " -DFALLBACK_TO_BYTES";
}

NV_MAX_PRECISION should be increased from 20 to 21 for long doubles where nvmantbits is 64.
Otherwise, uniqnum will regard some values that differ by 1ULP to be duplicates.
For example this version of  uniqnum() regards 10.770329614269008063 (0xa.c53452546cf9aa2) and 10.7703296142690080625 (0xa.c53452546cf9aa1) as duplicates.
Perhaps the test suite should test these values.

Similarly, I believe that for the case where nvmantbits is 112, we need NV_MAX_PRECISION (currently set at 35) to be 36 on rare occasions.
However, I haven't yet got an example where the value of 35 causes a failure.

And I still want to make a closer examination of what's happening with doubledoubles. (Again, no examples of any failure as yet.)

When nvsize is 8, I'm seeing NV_MAX_PRECISION being set to 20.
I can't think of any reason that it needs to be higher than 19, though setting it to 20 shouldn't break anything.
I see in the Makefile.PL a comment about wanting NV_MAX_PRECISION to at least match the precision of the UV (which is 20 decimal digits when ivsize == 8).
I altered the Makefile.PL to set NV_MAX_PRECISION to 19 and it didn't break anything for me, but going any lower than 19 does cause breakage,

That'll have to do for now as it's getting late.
First thing to establish is what to do regarding that issue with MS Windows.

Cheers,
Rob

sisyphus:
I've found some __float128 values that rely on NV_MAX_PRECISION being 36.

This current github version of S-L-U regards the two __float128 values of 0x1.f68d179b5295945c6f5aa46ff7c1p+9 and 0x1.f68d179b5295945c6f5aa46ff7c10p+9 (which differ by 1ULP) as being duplicates.

Similarly for:
0x1.fc4927a4a1dbff443353b0eff3b7p+9 & 0x1.fc4927a4a1dbff443353b0eff3b6p+9
and for:
0x1.fa2c8d9371ce11dd1b5d8c1ed76cp+9 & 0x1.fa2c8d9371ce11dd1b5d8c1ed76dp+9

AIUI, the correct calculation for determining the no. of decimal digits required is given by:
ceil(1 + (bits * log(2) / log(10)))

For bits values of 53, 64, and 113, this yields (respectively) 17, 21, and 36.

I forget the reason that we need 19 digits in our case when bits is 53. (Was it because of the 64-bit  IV/UV ?)

I'll have a look at the doubledoubles now.

Cheers,
Rob

sisyphus:
The "%.33g" formatting on my doubledouble build is appallingly buggy.
For example the "%.33g" formatting of 1+(2**-1073) and 1.00000000000000000000000000000005 results in the same string, and uniqnum() therefore decides that they are duplicates.
The 2 values are, however, miles apart.

It may well be that not all systems are as buggy as mine, but looking only at the most significant 33 digits is not a good basis for making comparisons, anyway.
I'd much prefer to be looking at the actual bytes, which works flawlessly.
I think I can unobtrusively build that in to this version of uniqnum.

My fix for the Windows-specific problem I mentioned earlier might not be so unobtrusive.
I'll see what I can do.

Seems to me that the code:

will read in the unused bytes when the NV is the extended precision (10-byte) long double.
Is it guaranteed that those unused bytes will not change from one NV to another ?
I'd feel more comfortable if those bytes weren't being read at all.

Cheers,
Rob

sisyphus:
I think I've got an acceptable version at https://github.com/sisyphus/Scalar-List-Utils
See what you think.
(I've made some alterations to ListUtil.xs, Makefile.PL and t/uniq.t.)

It's testing fine everywhere for me.
I'll create a new pull request tomorrow night after I've had a chance to chew on it a bit more.

In the meantime, thoughts/observations welcome.

Cheers,
Rob

haarg:
@sisyphus That patch series seems rather ugly and confusing.  It adds the FALLBACK_TO_BYTES section, then removes it.  The comments and commit messages all explain what the code is doing, but not why.  For example, nothing explains why `USED_NV_BYTES` is needed and why `sizeof(NV)` would be inaccurate.

sisyphus:
Yeah - a lot of those changes (including the addition and subsequent removal of FALLBACK_TO_BYTES) was done before I opened the PR.
I wasn't expecting the PR to present the entire history of the changes I had made to my clone prior to the opening of the PR.
This is a mistake that I'll avoid in future.

If you can compare my current ListUtil.xs with the current github version you should find it fairly straightforward.
I ditched FALLBACK_TO_BYTES when I realized I could work around that Windows problem by defining __USE_MINGW_ANSI_STDIO at the beginning of ListUtil.xs (unless perl's ccflags already defined it). 
That causes no alteration to your uniqnum() implementation, apart from a couple of small \#ifdef blocks.

I've also inserted a separate small block (\#ifdef NV_IS_DOUBLEDOUBLE) to cater for the doubledouble.

That's about it for my ListUtil.xs amendments.

My only concern with 'sizeof(NV)' is that (IIUC) the unused bytes of the 10-byte extended precision long double will be read in.
Can we be assured that those unused bytes will always be zero ?
If not, a difference in the unused bytes could make two identical long double values be deemed by uniqnum to be different.
There's a bit of an explanation at the start of the Makefile.PL.
I could certainly expand it a bit, and even put a comment about it in ListUtil.xs, too.
Of course, if someone likes to guarantee that we don't have to concern ourselves with this consideration then I'm quite happy to revert to 'sizeof(NV)'.

I wrote an Inline::C script to read all of the bytes of long doubles that were created in C space, and found that the first 2 of the unused bytes contained garbage on Windows.
Oddly, however, the garbage seemed to always be the same for every long double that was created - which doesn't cause any problem for uniqnum.
Disconcertingly, the garbage changed from one run to the next - but the garbage remained consistent throughout for each run.
IOW, I can't produce any failure caused by 'sizeof(NV)' ... I'm just too faint-hearted (and lacking in certainty) to assert that it won't ever cause a problem.

I think the dumbest thing I did was to initially change your NV_MAX_PRECISION setting for doubles from 20 to 19 - which I did because the test script didn't complain when I made the switch to 19.
It was dumb because I knew that your reasoning for choosing 20 sounded logical.
Eventually I knuckled down, found some values that fail for 19, amended t/uniq.t to include tests that expose the fact, and changed the value back to 20.

In summary, I reckon the net changes I've made compared to the current github version are not extensive.

And if someone wants to tell me not to bother about the unused bytes, then I think we could actually leave the current github Makefile.PL unchanged.
Just say the word and I'll do that.

Just one last thought on your ListUtil.xs (which I've just been looking over again):
It seems to be geared to handle the case of ivsize == nvsize == 8 very nicely.
I have the notion that this code of yours should be limited to handling that single case, and every other configuration should be handled by the code that's currently in my NV_IS_DOUBLEDOUBLE block. 
All that's required is to change:
\#ifdef NV_IS_DOUBLEDOUBLE
to
\#ifdef NVSIZE_NOT_EQUAL_IVSIZE
with NVSIZE_NOT_EQUAL_IVSIZE being defined by the Makefile.PL whenever $Config{nvsize} != $Config{ivsize}

I'm assuming that:
sv_setpvn(keysv, (char \*) &nv_arg, sizeof(NV));
is less expensive than:
sv_setpvf(keysv, "%0.\*" NVgf, NV_MAX_PRECISION, nv_arg);

Also, the appending of the "f" to keysv should not be needed when ivsize != nvsize .

That would mean NV_MAX_PRECISION is hard coded to 20, because no other configuration needs to make use of NV_MAX_PRECISION.

(I probably should test this before getting too carried away ... something for tomorrow ;-)

Thanks for the feedback.
Looking forward to hearing more from you. 

Cheers,
Rob

PR #84, opened 2019-12-04
another attempt to fix uniqnum()

sisyphus:
Changes made:
1) Change the set of values that NV_MAX_PRECISION can take from 20, 20 and 35 to 19, 21 and  36 (respectively); 
2) When reading the NV's bytes, read only those bytes that are actually used by the NV, and ignore the unused ones;
3) Accommodate doubledouble NV's reliably;
4) Accommodate Win32 perls whose formatting of NVs is buggy for certain values.

Note that the last of those changes provides a fix only for Win32 perls built using a mingw port of gcc.
For Win32 perls built using a Microsoft compiler, this particular problem (which has existed since the inception of the uniqnum function) is probably not fixed.

Note also that the uniqnum code for the doubledouble, which is quite compact, could also be used for all configurations of perl where $Config{nvsize} > $Config{ivsize}.
(However, given the need to also accommodate the case where $Config{nvsize} and $Config{ivsize} are both 8, I'm not sure that there's any advantage in making use of that.)

Cheers,
Rob


sisyphus:
Haarg's reasoning for setting NV_MAX_PRECISION to 20 when ivsize == nvsize looked sound to me.
But I couldn't find any exampls where an NV_MAX_PRECISION of 19 would not suffice.

I have now found them - so I'll need to make some minor alterations to Makefile.PL.
I'll also amend t/uniq.t to include a test that exposes the need for NV_MAX_PRECISION to be set to 20.

I'll do that today or tomorrow.

Cheers,
Rob

sisyphus:
This latest commit rearranges things in a way that, I believe, improves simplicity and performance.
To enable people to better gauge the changes, I'm attaching a diff that compares the current state of my sisyphus/Scalar-List-Utils clone to the current Dual-Life/Scalar-List-Utils.

Hopefully, this is getting closer to something that can be accepted by all.

Cheers,
Rob

[SLU.diff.txt](https://github.com/Dual-Life/Scalar-List-Utils/files/3930148/SLU.diff.txt)

karenetheridge:
> To enable people to better gauge the changes, I'm attaching a diff

That can be seen right from the pull request, in the "files changed" tab: https://github.com/Dual-Life/Scalar-List-Utils/pull/84/files

sisyphus:
I think the layout that diff provides for the Makefile.PL changes in the attachment is clearer than that provided by github.
For the other 2 files, it's much of a muchness - the layouts may even be identical.
Of course, I didn't know where to find those diffs on gitbub so thanks for the info.

I've again reached that point where I'm thinking there's nothing more for me to do.
If I'm mistaken, please provide some instruction on what else is needed.
Otherwise, what are we now waiting for ?

Cheers,
Rob

sisyphus:
Why has there been no action on this in the last 10 days ?
Or has there been some action that I've missed ?

Cheers,
Rob

ListUtil.xs
```diff
@@ -1238,15 +1256,46 @@ CODE:
...
                  * a UV or IV */
                 else {
-                    sv_setpvf(keysv, "%0.*" NVgf, NV_MAX_PRECISION, nv_arg);
+#ifdef WIN32_PERL_NO_ANSI
+
+                   /* Because perl was not built with ansi compliance, doing: *
+                    * sv_setpvf(keysv, "%0.20" NVgf, nv_arg)                  *
+                    * will not always work as intended.                       *
+                    * But the following workaround does what we want.         */
+
```
leonerd: As it's the first statement inside the `{` block, the declaration of `buffer` could be made here, inside the `#ifdef`, reducing its scope to where it is used and overall neater

sisyphus: Cool .... and done.
All tests still passing.

leonerd:
Overall, this is still a large set of changes to review and think about. I'd suggest two things:

1. You list four separate changes in your initial "list of changes". Could you instead submit four separate issues, one for each of them
2. Could you approach each of them, not from the perspective of what change you are proposing to make, but what problem it will fix. That will help us understand the overall context of *why* the changes are being proposed, which currently is rather hard to decipher given only the solution to it.

Also in general with more foreknowledge of the problem being addressed, it may be that we can suggest a different fix that would be more suitable. A better plan of approach would be to first raise an RT ticket to simply explain what the problem is and begin discussing it, before looking into a code implementation to fix it.

sisyphus:
In the github version of List::Util's uniqnum implementation we have:
/* for numbers outside of the IV or UV range, we don't need to
 * use a comparable format, so just use the raw bytes, adding
 * 'f' to ensure not matching a stringified number */
 else if (nv_arg < (NV)IV_MIN || nv_arg > (NV)UV_MAX) {
   sv_catpvn(keysv, "f", 1);
 }

I'm fairly confident that we don't need that sv_catpvn() call.
I can't find any NVs less than IV_MIN or greater than UV_MAX whose raw bytes will match a stringified number,
In fact, all of the NVs (that I've checked) whose raw bytes do look like a stringified number have an absolute value of less than 1.

Here's the search that I conducted - looking for values greater than 1, and finding none:

use warnings;

for(10000000 .. 99999999) {
 if(abs(unpack('d', pack('a8', "$_")))  > 1)
   { print "$_\n" }
}

for(1000000 .. 9999999) {
  substr($_, 1, 0, '.');
  if(abs(unpack('d', pack('a8', "$_")))  > 1)
    { print "$_\n" }
}

for(1000000 .. 9999999) {
  substr($_, 6, 0, '.');
  if(abs(unpack('d', pack('a8', "$_")))  > 1)
    { print "$_\n" }
}

for(100000 .. 999999) {
  substr($_, 4, 0, 'e+');
  if(abs(unpack('d', pack('a8', "$_")))  > 1)
    { print "$_\n" }
}

for(100000 .. 999999) {
  substr($_, 4, 0, 'E+');
  if(abs(unpack('d', pack('a8', "$_")))  > 1)
    { print "$_\n" }
}

for(100000 .. 999999) {
  substr($_, 4, 0, 'e-');
  if(abs(unpack('d', pack('a8', "$_")))  > 1)
    { print "$_\n" }
}

for(100000 .. 999999) {
  substr($_, 4, 0, 'E-');
  if(abs(unpack('d', pack('a8', "$_")))  > 1)
    { print "$_\n" }
}

for(1000000 .. 9999999) {
  if(abs(unpack('d', pack('a8', "-" . "$_")))  > 1)
    { print "$_\n" }
}

for(1000000 .. 9999999) {
  if(abs(unpack('d', pack('a8', "+" ."$_")))  > 1)
    { print "$_\n" }
}
__END__
Produces no output.

Cheers,
Rob

sisyphus:
This is in response to @leonerd's post of 4 days ago.

I've always thought it's best to start with something that works, and then work on optimizing it.

For the purposes of simplicity, let's call the configuration that sets both $Config{ivsize} and $Config{nvsize} to 8 "the nuisance configuration".
This, of course, is the most common configuration of perl these days.

For every other configuration (all of which, importantly, set nvsize to be larger than ivsize), a straightforward uniqnum() implementation can be coded quite easily as follows:

Given an SV * arg we just do:

\======================================
NV nv_arg = SvNV(arg);
if(nv_arg == 0) sv_setpvs(keysv, "0");
else if(nv_arg != nv_arg) sv_setpvf(keysv, "%" NVgf, nv_arg);
else sv_setpvn(keysv, (char *) &nv_arg, USED_NV_BYTES);
\======================================

That's the approach I've taken - and it's the first point of divergence between my approach and the approach taken by both 1.53 and current git. 
Can we agree that this is a sensible approach ?
It's clean and simple, and works flawlessly for every non-nuisance configuration of perl.
And it leaves us free to concentrate on the nuisance configuration, which is where the portability issues arise.
Sure, it means converting IVs and UVs to NVs, but then you only have to read in the bytes - you don't have to stringify the values, as 1.53 and git currently want to do.
(I'll mention here that with double-doubles, basing anything on the stringification of values is flawed, whereas looking at the byte structure is reliable.)

If we can agree on that, then I'll happily provide a version of uniqnum that works flawlessly on linux - which is something that can't be said for either 1.53 or current git.
I can leave out the Windows-specific fixes (for now) if you want - but if you also include (as you ought) an open and honest version of uniqnum.t (as opposed to the current version that conceals the issues) there will be reports of failures for Windows builds prior to version 5.26, and for Windows perls built with Microsoft compilers.

I've actually got what I think is a pretty reasonable fix for the issue with Microsoft compilers - though it hasn't yet been pushed to my fork of Scalar-List-Utils.
See Uniqnum.xs in List-Uniqnum-0.04 on CPAN. (Look for _MSC_VER which is the symbol that identifies MS compilers.)

And if you want to see exactly what's wrong with uniqnum() in 1.53 or github, just grab:

https://metacpan.org/source/SISYPHUS/List-Uniqnum-0.04/t/01uniqnum.t

and change line 7 from:

use List::Uniqnum qw( uniqnum );
to
use List::Util qw( uniqnum );

and see what happens when you run it.
(I'm quite happy to enter into a discussion about the results.)

I see that there have been some changes to the github version of Scalar::List::Utils since I last checked.

What follows relates to this current github version:

I can report that, with perl-5.31.8 on Windows, the nuisance configuration passes all tests with it's own t/uniq.t and my own 01uniqnum.t.
It's the same with the nuisance configuration of 5.30.0 on Linux.

With perl-5.31.8 on Windows, a non-nuisance configuration (-Duselongdouble) fails tests 5-6 with my 01uniqnum.t, but passes all t/uniq.t.
It's the same with my -Duselongdouble build of perl-5.30.0 on Linux. (The t/uniq.t script is at fault here, because it is hiding an uniqnum inadequacy.)
The problem is that uniqnum is claiming that 0xa.c53452546cf9aa2p+0 and 0xa.c53452546cf9aa1p+0 are duplicates, even though they differ by one unit in the last place.
The former is an exact representation of the rational 12417344623994706594 / 1152921504606846976.
The latter is an exact representation of the rational 12417344623994706593 / 1152921504606846976.
Many would say that the 2 values are pretty close to each other.
However, they are not duplicates by any stretch of the imagination.
The same type of issue arises with my -Dusequadmath build of perl-5.30.0 on Linux.
In this case uniqnum() incorrectly regards
0x1.f68d179b5295945c6f5aa46ff7c1p+9
and
0x1.f68d179b5295945c6f5aa46ff7cp+9
as duplicates.


With perl-5.24.0 on Windows, the nuisance configuration fails test 19 in t/uniq.t and tests 8,25 in my 01uniqnum.t.
These failures are the same issue - an issue that afflicts only the nuisance configuration.
It's a Windows-specific issue that was addressed by my pull request.
It's a failure to recognize that (eg) the IV 100000000000262144 and the NV (double) 100000000000262144.0 provide exact representations of the same value and are therefore duplicates of each other.
It's the problem with "%.20g" formatting when perl was built without -D__USE_MINGW_ANSI_STDIO.
It's also the same problem as faced with MS compilers - except that, for the MS compilers, I've had to find a different workaround.

On my nuisance configuration of perl-5.16.0 built with a Microsoft compiler (cl version 14.00.40310.41), the github version of Scalar::List::Utils won't even compile:

ListUtil.obj : error LNK2001: unresolved external symbol trunc
blib\arch\auto\List\Util\Util.dll : fatal error LNK1120: 1 unresolved externals

It's possible that more recent Microsoft compilers provide "trunc", but I figure that if you want to support perl-5.6 then you'll probably want to support this MS compiler (which I think is newer than perl-5.6).
I worked around the problem in the obvious way, by changing

ret -= trunc(ret);
to
ret -= ret > 0 ? floor(ret) : ceil(ret);

If you wanted to limit the change to MS compilers only:

#ifdef _MSC_VER
ret -= ret > 0 ? floor(ret) : ceil(ret);
#else
ret -= trunc(ret);
#endif

With that done, I get the same failures as I got with the nuisance configuration of perl-5.24.0 - namely test 19 of uniq.t fails, as do tests 8 and 25 in 01uniqnum.t.

BTW, my sole reason for releasing List::Uniqnum to cpan is to obtain broader based testing results via cpantesters.

Cheers,
Rob

sisyphus:
Looking at 1.54, I can see that I've been stonewalled again.
I'll therefore close this issue, and create a new one soonish.
It will be a minimal PR that simply fixes the only remaining issue with List::Util on Linux that I know of.
That issue is that on -Duselongdouble and -Dusequadmath builds of perl, uniqnum() still wants to claim that some unique (unequal) floating point values are duplicates of each other.
This is happening because NV_MAX_PRECISION is being set to 1 lower than it ought to be for those 2 configurations of perl - and it's a bug in t/uniq.t that's preventing the failure from being detected.
I'll leave the Windows issues alone until we get Linux fixed.

Cheers,
Rob

PR #85, opened 2020-01-28
Customisable rand() for shuffle() [RT#131536]

leonerd:
Fixes https://rt.cpan.org/Ticket/Display.html?id=131536


dur-randir:
'randcv' shouldn't change mid-flight, yet you check it on every iteration.

leonerd:
@dur-randir  Yes; it avoids the code duplication of having two entire copied bodies of the loop. I expect the overall simplicity outweighs any tiny advantage of runtime, especially as the rest of the code is running quite a lot of other operations. If you think a copy-paste approach would be much faster, I'll await a benchmark result to see.

PR #86, opened 2020-01-29
Add List::Util::sample() (RT131535)

leonerd:
Resolves https://rt.cpan.org/Ticket/Display.html?id=131535

### DIFF
--- a/ListUtil.xs
+++ b/ListUtil.xs
@@ -1206,7 +1206,7 @@ CODE:
 
 void
 sample(...)
-PROTOTYPE: @
+PROTOTYPE: $@
 CODE:
 {
     UV count = items ? SvUV(ST(0)) : 0;
--- a/ListUtil.xs
+++ b/ListUtil.xs
@@ -1204,6 +1204,52 @@ CODE:
     XSRETURN(items);
 }
 
+void
+sample(...)
+PROTOTYPE: @
+CODE:
+{
+    UV count = items ? SvUV(ST(0)) : 0;
+    int reti = 0;
+    SV *randsv = get_sv("List::Util::RAND", 0);
+    CV * const randcv = randsv && SvROK(randsv) && SvTYPE(SvRV(randsv)) == SVt_PVCV ?
+        (CV *)SvRV(randsv) : NULL;
+
+    if(!count)
+        XSRETURN(0);
+
+    /* Now we've extracted count from ST(0) the rest of this logic will be a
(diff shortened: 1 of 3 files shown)


lib/List/Util.pm
```diff
@@ -494,6 +495,51 @@ Returns the values of the input in a random order
...
+
+  my @ret;
+  while( @_ and $count-- ) {
+    my $i = $List::Util::RAND ?
+      int( _ffrac( $List::Util::RAND->() ) * @_ ) :
+      int rand @_;
+    push @ret, $_[$i];
+
+    # delete it from the list
+    $_[$i] = $_[-1] if $i < $#_; # No point doing this if $i is the final item
```
wolfsage: This modifies the values of the array passed in. `sample 1, @foo` will change `@foo`

leonerd: Ughyes, you're right. A fix for that which still retains any kind of performance will be nontrivial. I guess a rewrite into XS is called for.

Now done at 624a150

ListUtil.xs
```diff
@@ -1204,6 +1204,52 @@ CODE:
...
 
+void
+sample(...)
+PROTOTYPE: @
+CODE:
+{
+    UV count = items ? SvUV(ST(0)) : 0;
+    int reti = 0;
+    SV *randsv = get_sv("List::Util::RAND", 0);
+    CV * const randcv = randsv && SvROK(randsv) && SvTYPE(SvRV(randsv)) == SVt_PVCV ?
```
wolfsage: For performance if count is 0 (silly) might be better to grab randcv below `if (!count) XSRETURN(0)` but seems rare enough to probably not be worth it.

ListUtil.xs
```diff
@@ -1204,6 +1204,52 @@ CODE:
     XSRETURN(items);
 }
 
+void
+sample(...)
+PROTOTYPE: @
```
haarg: I'd say this should have a prototype of `$@` for consistency with head and tail.

ListUtil.xs
leonerd: That's possible but then (because C89) the variable needs declaring on a separate line to assignment. Not worth the readability hit for that tiny performance improvement in a rare case.

ListUtil.xs
leonerd: Yeah; probably best. Annoyingly that means I can't use the `ALIAS` mechanism to share the implementation with `shuffle` any more, but fair enough.

Done 56d89f2

PR #87, opened 2020-01-31
Add List::Util::reductions

leonerd:
As requested in https://rt.cpan.org/Ticket/Display.html?id=128237

### DIFF
--- a/ListUtil.xs
+++ b/ListUtil.xs
@@ -515,11 +515,12 @@ CODE:
     if(cv == Nullcv)
         croak("Not a subroutine reference");
 
-    if(items <= 1)
+    if(items <= 1) {
         if(ix)
             XSRETURN(0);
         else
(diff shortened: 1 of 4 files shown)


dur-randir:
- it generates new compiler warning
- it leaks if code block dies

leonerd:
@dur-randir Ahyes; the leak I can imagine where that comes from and I think I have an idea on fixing it. But what new warning is generated? I don't get anything from gcc - though it quite often doesn't complain.

leonerd:
496d642 now handles destruction of accumulated temporaries after exceptions.

Still awaiting more detail on what the compiler warning is

leonerd:
I believe this may be ready. @dur-randir any comment?

dur-randir:
Since you already use AvARRAY, you could also use it to push items in the ST() manner, but otherwise it's fine.

PR #88, opened 2020-02-03
remove most tabs from source

haarg:
Most of the code was written using spaces for indentation, so convert
most of the rest to use spaces as well.  Tabs were used in some places
for alignment, not indentation, and even there it was very inconsistent.

Some files were left alone, such as multicall.h because it mostly
consistently uses tabs, and Changes to leave its historic content in
place.


leonerd:
An excellent idea. I keep meaning to get around to doing that but you beat me to it :)
Thanks.

PR #89, opened 2020-02-03
avoid done_testing for compatibility with old perl

haarg:
done_testing was added in Test::More 0.88, and isn't available on stock
perls older than 5.10.1.  Adding a dependency on a newer Test::More
would introduce a dependency loop on those older perls.


leonerd:
LGTM

PR #90, opened 2020-02-03
test predictable $RAND giving consistent results, not predefined results

haarg:
Rather than testing the output of shuffle or sample with a predictable
$RAND resulting in a specific pattern, run the function multiple times
and verify it gives the same results.  This means the test doesn't have
to rely on the specific implementation of the functions, since there are
many ways they could be written.  Users also shouldn't be relying on a
given random function giving identical results across versions or
systems.


leonerd:
LGTM

PR #91, opened 2020-02-04
Fix uniqnum for perls whose nvsize is greater than 8

sisyphus:
For perls whose nvsize is greater than 8 the XS symbol NV_MAX_PRECISION was being set too low (by 1).
This pull request fixes that (in the Makefile.PL, where the value of the symbol is determined).
Tests 16 and 17 of t/uniq.t were intended to catch this failure, but failed to do so owing to the particular values that were chosen for those tests.
Therefore, those tests have also been corrected such that they would have caught the error.
If this amended t/uniq.t is run against List-Util-1.54 on a perl whose nvsize is greater than 8, you will find that tests 16 ('uniqnum preserves uniqueness of high precision floats') and 17 ('uniqnum preserves uniqueness of high precision floats (stringified))' fail.

Cheers,
Rob


haarg:
This change is probably fine, but it would really be nice if the comment explained why the calculation is not correct.  The current comment just states what the code does, not why.

sisyphus:
Let's see ... there's some code (that I find hard to understand) in the 1.54 Makefile.PL that doesn't come up with the correct result.
I add one line to it so that it does come up with the correct result - and now it's up to *me* to explain what that code did, where it went wrong, and why it needed fixing. (Of course, it needed fixing simply because it was wrong.)
*I* think the person that wrote that code in the first place (and presumably understands the logic of it) should do the explaining.

Hint:
The value we're looking for (as I've written previously) is ceil(1+ log(2) / log(10) *  prec) where "prec" is the precision (in bits) of the NV's mantissa (including the leading implicit bit, where applicable).
Ignoring the doubledouble (for which NV_MAX_PRECISION is unworkable), those values will be 17 for nvsize == 8, 21 for the 80-bit long double, and 36 for IEE long double and __float128.
The one and only caveat to that is for that one case where ivsize == nvsize, when we need a value of 20.

Come up with something that does that, and you've nailed it.
Something like the following would be fine:

if( $Config{nvsize} == $Config{ivsize} ) { $nv_digits = 20 }
elsif( $Config{nvsize} == 8 ) { $nv_digits = 17 }
elsif( length(sqrt 2) < 30 ) { $nv_digits = 21 }
else{ $nv_digits = 36 }
$defines .= " -DNV_MAX_PRECISION=$nv_digits";

or, if you wanted to provide some insight into the way those values were derived:

\# Generally NV_MAX_PRECISION needs to be set  to ceil(1 + (log 2 / log(10) * prec))
\# where prec is the bit-precision of the NV's mantissa, including any implied bits.
\# However, if nvsize <= ivsize NV_MAX_PRECISION needs to be set to the length
\# of UV_MAX (where UV_MAX is the UV with the greatest magnitude).
if( $Config{nvsize} <= $Config{ivsize} ) { $nv_digits = length ~0 }
elsif( $Config{nvsize} == 8 ) { $nv_digits = POSIX::ceil(1 + (log 2 / log(10) * 53)) }
elsif( length(sqrt 2) < 30 ) { $nv_digits = POSIX::ceil(1 + (log(2) / log(10) * 64)) }
else{ $nv_digits = POSIX::ceil(1 + (log(2) / log(10) * 113)) }
$defines .= " -DNV_MAX_PRECISION=$nv_digits";

If any of that is deemed to be acceptable, then I'll push it to my S-L-U fork. (I shall, of course, first check that it works as intended.)

BUT !!
If uniqnum() was implemented the way I think it should be, then the calculation of NV_MAX_PRECISION becomes completely unnecessary. 

My uniqnum implementation:
For the case where ivsize == nvsize, I would hard code in the value 20.
For all other configurations I would just read in the bytes. (When we read in the bytes, and therefore don't stringify values, we don't need to determine the value of NV_MAX_PRECISION.)
That's the approach I took in PR 84, and in List::Uniqnum.

Of course, to read in the bytes, we *do* need to know how many of them there are.
For that we can use (as you've suggested) "sizeof(NV)".
My only objection to "sizeof(NV)" is that, in the case of the 80-bit extended precision "long double", you're reading in more bytes than are needed - sizeof(NV) will be 12 or 16, but we need look at only the first 10 bytes.
And if any of the unused bytes contain arbitrary values, then we definitely need to ignore them.
Hence, I would prefer to define a USED_NV_BYTES value in Util.xs.
Its value is set to $Config{nvsize} unless the nvtype is the 80-bit extended precision long double, in which case USED_NV_BYTES is set to 10.
That's very easy for the Makefile.PL to achieve:

$nvsize = $Config{nvsize};
if($nvsize > 8 && $Config{nvtype} eq 'long double') {
  $nvsize = 10 if length sqrt 2.0 < 30;
}
$defines .= " -DUSED_NV_BYTES=$nvsize";

We *could* detect the 80-bit long double using $Config{longdblkind} instead of length(sqrt 2).
However, there's a portability issue with the use of $Config{longdblkind} in that it hasn't been around for all that long.

BTW, have you noticed the failures of 1.54 with a couple of Windows smokers ?
They arise because those builds of perl did not define __USE_MINGW_ANSI_STDIO.
I've harped on about that previously, too ... and provided a fix for that problem in PR 84.

Anyway ... there goes another 2 hours of my life spent responding to an objection that took a couple of minutes to compose.

Cheers,
Rob

sisyphus:
Just to explain a little better, @leonerd asked me in his last post to PR 84 ( https://github.com/Dual-Life/Scalar-List-Utils/pull/84 ) to present my changes bit by bit, rather than all at once.
So, that's what I'm trying to do and hence the minimalist approach of my initial changes.
I expect we'll eventually end up with something very like PR 84, but ... we'll just have to wait and see.

I've just realized that the subject line I've chosen for this PR is misleading because it doesn't cater for the doubledouble.
I'll deal with doubledoubles next - after we've dealt with the particular issue this PR is addressing.
It will require a a coupla lines of code in the Makefile.PL to detect that NV is doubledouble, and its own #ifdef block in uniqnum to handle the case.
In my PR 84 it was unnecessary to do any of that - the doubledouble just gets automatically handled correctly without any need to detect it.

Cheers,
Rob

leonerd:
LGTM

Grinnz:
@sisyphus  I appreciate you taking the time to explain this change, and to contribute the solution, and sorry you feel snubbed by your changes languishing for so long. I am not personally knowledgeable to discuss these changes, but speaking as a volunteer as we all are, it is much more difficult to evaluate such detailed changes and their intended impact based on only the code itself, as well as when each logically separate change can be evaluated on its own. So I hope you understand why such requests were made so your changes can be better evaluated.

sisyphus:
@Grinnz most of the delays were warranted, resulting in corrections and worthwhile improvements - especially @haarg's introduction of sv_setpvn() and friends to replace my farting about with C buffers.
.
It was mainly the lack of feedback from testers that has been frustrating me - and then I was further deflated when 1.54 was released with a test script that still hid the problem that this merge addresses.
That was when I released List::Uniqnum just so I could verify that the approach I envisaged was working, beyond the restricted  spectrum of my own systems and builds of perl.

The merging of this PR has significantly buoyed my spirits, and I'm ready to move on to the next issue.
I said above that I would next like to fix uniqnum to handle doubledouble NVs - but I've decided to instead address that Windows issue that is causing the 3 Windows failures on the cpantesters matrix.
I think it's probably best to deal first with the fail reports from the testers. (AFAIK there are no smokers testing doubledouble builds.)

I'll open another PR for that Windows issue as soon as I've checked that I've got it right.
That will probably be tomorrow.

Thanks guys !!

Cheers,
Rob

PR #93, opened 2020-02-04
avoid using Scalar::Util functions in List::Util tests

haarg:
This ensures that changes to one won't impact the test for the other.
It also prepares for a potential future where the modules are split into
separate dists.  This also makes it easier to port tests to
List::Util::PP, which wants to avoid all direct dependencies on
Scalar::Util.

This change is admittedly mainly to make it easier for me to maintain List::Util::PP.


leonerd:
I'm happy with the idea, though I'm a little worried how we'll maintain this going forward. Perhaps put a comment at the top of each .t file where the `use Scalar::Util ...` line used to be, to say that we're not using S:U and why.

PR #94, opened 2020-02-06
move Test::More to be a test dependency

haarg:
Properly list Test::More as a test dependency rather than a runtime
dependency.  This requires adding some code to cope with older versions
of EUMM that don't understand TEST_REQUIRES.


leonerd:
LGTM

PR #96, opened 2020-02-08
Improve uniqnum's handling of configurations where IVSIZE != NVSIZE

sisyphus:
In ListUtil.xs, I inserted an "#if IVSIZE != NVSIZE" block into uniqnum().
That block is needed for the case where NV is doubledouble - but it so happens that all other configurations where IVSIZE != NVSIZE" are handled more efficiently by that same approach.
Therefore, rather than create that block as "#ifdef NV_IS_DOUBLEDOUBLE",  I've expanded it to handle all cases where IVSIZE != NVSIZE.
At the end of this post, I'll include the benchmark script which (hopefully) demonstrates the benefit of this approach.

Having done that, it then meant that the one occurrence of NV_MAX_PRECISION was always 20.
Hence, we're able to replace that occurrence of NV_MAX_PRECISION with a hard-coded "20", and remove the calculations of NV_MAX_PRECISION from the Makefile.PL.

I suppose one annoyance with this is that there are 2 different paths we can take in uniqnum that result in performing essentially the same procedure of check for 0, check fr NaN, read the bytes into keysv.
The first place is  in the "#if IVSIZE != NVSIZE" block,
The second place is in a section of the accompanying "#else" block.

I'm not sure whether that doubling up can be eliminated - something to look at next, perhaps.
For the moment, I just wanted to keep the logic set out as clearly as possible.

Here's the benchmark comparison script I used:

\=============================
use warnings;
use Benchmark;
use Inline C => <<'EOC';

void as_iv (SV * in, ...) {
  dXSARGS;
  int i;
  SV *keysv = sv_newmortal();

  for(i = 0; i < items; i++) {
    if(!SvOK(ST(i)) || SvUOK(ST(i))) {
      sv_setpvf(keysv, "%" UVuf, SvUV(ST(i)));
    }
    else if(SvIOK(ST(i))) {
      sv_setpvf(keysv, "%" IVdf, SvIV(ST(i)));
    }
    else {
      croak("Should not have got to here");
    }
  }
}

void as_nv(SV * in, ...) {
  dXSARGS;
  int i;
  NV nv_arg;
  SV *keysv = sv_newmortal();

  for(i = 0; i < items; i++) {
    nv_arg = SvNV(ST(i));

    if(nv_arg == 0) {
      sv_setpvs(keysv, "0");
    }
    else if (nv_arg != nv_arg) {
      sv_setpvf(keysv, "%" NVgf, nv_arg);
    }
    else {
    }
  }  
}

EOC

my @uv;
my $iv_high = ~0 >> 1;
my $iv_low  = $iv_high - 90000;

for($iv_low .. $iv_high) { push @uv, $_ << 1; }

@iv = (1 .. 100000, 900000 .. 1000000,
       100000000 .. 100090000,
       1000000000 .. 1000090000, @uv);

print scalar @iv, "\n";

die "Input array not as expected" unless @iv == 470004;

timethese (1, {
  'AS_IV' => 'as_iv(@iv);',
  'AS_NV' => 'as_nv(@iv);',
});
\=============================

The output for my -Duselongdouble build of perl on Windows is:

Benchmark: timing 1 iterations of AS_IV, AS_NV...
     AS_IV:  0 wallclock secs ( 0.08 usr +  0.00 sys =  0.08 CPU) @ 12.82/s (n=1)
     AS_NV:  0 wallclock secs ( 0.02 usr +  0.01 sys =  0.03 CPU) @ 32.26/s (n=1)

Irrespective of the configuration of perl, that script always claims that as_nv() is at least twice as fast as as_iv().

I don't think I've skewed the testing in favour of one approach over the other.

I'll just repeat that distinguishing doubledoubles on the basis of stringified values is unworkable. For doubledoubles we really do need to look at the byte structure.
(More than happy to elaborate on that, if requested.)

I presume travis will check the IVSIZE==NVSIZE configuration thoroughly.
I have tested most of the IVSIZE!=NVSIZE configurations - including nvtypes of 80-bit long double and __float128 on my local windows and linux systems.
I've also tested the case where IVSIZE is 4 && NVSIZE is 8.
And, of course, I've tested the doubledouble.

Cheers,
Rob

### DIFF
--- a/ListUtil.xs
+++ b/ListUtil.xs
@@ -1349,7 +1349,11 @@ CODE:
                 SvNV(arg); /* SvIV() sets SVf_IOK even on floats on 5.6 */
 #endif
             }
-#if IVSIZE != NVSIZE  /* open nvsize!=ivsize block */
+#if NVSIZE > IVSIZE  /* open nvsize>ivsize block */
+
+        /* We can detect duplicates by looking *
+         * solely at the value of SvNV(arg)    */
+              
         nv_arg = SvNV(arg);
 
         if(nv_arg == 0) {
@@ -1364,7 +1368,11 @@ CODE:
             /* Use the byte structure of the NV. */
             sv_setpvn(keysv, (char *) &nv_arg, sizeof(NV));  
         }
-#else                 /* close ivsize!=nvsize block, open ivsize==nvsize block */
+#else                 /* close nvsize>ivsize block, open ivsize==nvsize block */
+
+            /* Because IVSIZE == NVSIZE we cannot rely *
+             * solely on the value of SvNV(arg)        */
+
             if(!SvOK(arg) || SvUOK(arg)) {
                 sv_setpvf(keysv, "%" UVuf, SvUV(arg));
             }
--- a/ListUtil.xs
+++ b/ListUtil.xs
@@ -1349,7 +1349,22 @@ CODE:
                 SvNV(arg); /* SvIV() sets SVf_IOK even on floats on 5.6 */
 #endif
             }
+#if IVSIZE != NVSIZE  /* open nvsize!=ivsize block */
+        nv_arg = SvNV(arg);
 
+        if(nv_arg == 0) {
+            /* use 0 for both 0 and -0.0 */
+            sv_setpvs(keysv, "0");
+        }
+        else if (nv_arg != nv_arg) {
+            /* for NaN, use the platform's normal stringification */
+            sv_setpvf(keysv, "%" NVgf, nv_arg);
+        }
+        else {
+            /* Use the byte structure of the NV. */
+            sv_setpvn(keysv, (char *) &nv_arg, sizeof(NV));  
+        }
+#else                 /* close ivsize!=nvsize block, open ivsize==nvsize block */
             if(!SvOK(arg) || SvUOK(arg)) {
                 sv_setpvf(keysv, "%" UVuf, SvUV(arg));
             }
@@ -1376,7 +1391,7 @@ CODE:
                 /* smaller floats get formatted using %g and could be equal to
                  * a UV or IV */
                 else {
-#if defined(MINGW_PERL_NO_ANSI) && NV_MAX_PRECISION == 20
+#if defined(MINGW_PERL_NO_ANSI)
 
                    /* Because perl was not built with ansi compliance, doing: *
                     * sv_setpvf(keysv, "%0.20" NVgf, nv_arg)                  *
@@ -1386,10 +1401,11 @@ CODE:
                     sprintf(buffer, "%0.20" NVgf, nv_arg);
                     sv_setpvf(keysv, "%s", buffer);
 #else
-                    sv_setpvf(keysv, "%0.*" NVgf, NV_MAX_PRECISION, nv_arg);
+                    sv_setpvf(keysv, "%0.20" NVgf, nv_arg);
 #endif
                 }
             }
+#endif             /* close ivsize==nvsize block */
 #ifdef HV_FETCH_EMPTY_HE
             he = (HE*) hv_common(seen, NULL, SvPVX(keysv), SvCUR(keysv), 0, HV_FETCH_LVALUE | HV_FETCH_EMPTY_HE, NULL, 0);
             if (HeVAL(he))
--- a/Makefile.PL
+++ b/Makefile.PL
@@ -8,21 +8,6 @@ use ExtUtils::MakeMaker;
 my $PERL_CORE = grep { $_ eq 'PERL_CORE=1' } @ARGV;
 my $defines = $ENV{PERL_CORE} ? q[-DPERL_EXT] : q[-DPERL_EXT -DUSE_PPPORT_H];
 
-my $nv_digits;
-
-# Generally NV_MAX_PRECISION needs to be set to ceil(1 + (log 2 / log(10) * prec))
-# where prec is the bit-precision of the NV's mantissa, including any implied bits.
-# However, if nvsize <= ivsize NV_MAX_PRECISION needs to be set to the length
-# of UV_MAX (where UV_MAX is the UV with the greatest magnitude).
-
-if( $Config{nvsize} <=
-    $Config{ivsize} )         { $nv_digits = 20 } # length(~0) 
-elsif( $Config{nvsize} == 8 ) { $nv_digits = 17 } # POSIX::ceil(1 + (log(2) / log(10) *  53))
-elsif( length(sqrt 2) < 30 )  { $nv_digits = 21 } # POSIX::ceil(1 + (log(2) / log(10) *  64))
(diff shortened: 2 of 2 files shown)


sisyphus:
Added a couple of lines of documentation.
Also changed the condition "IVSIZE != NVSIZE" to "NVSIZE > IVSIZE".
Effectively they're the same condition because NVSIZE is never less than IVSIZE.
But the key to being able to detect duplicates by looking solely at SvNV(arg) is that NVSIZE is greater than IVSIZE.

sisyphus:
So ... what's the problem with this PR ?

Cheers,
Rob

sisyphus:
I guess that when someone asks "what's the problem with this PR ?", and there is no problem, then there's really no need for anyone to reply.
That's fair enough - strictly speaking, the question cannot be answered if there is no problem, and therefore it warrants no response.

However, if that question was based upon a fallacious presupposition, then I'm left thinking "why hasn't this PR already been merged ?"
(And perhaps someone could then answer that latter question.)

Did I take too many steps in this PR ?
If needed, I'll close this PR and open a new one that just takes one little baby step of enabling doubledouble support - and nothing else.

Or is it just that no-one here cares that this current github implementation of uniqnum() is broken for the "doubledouble" nvtype ?

In current github, we now have a uniqnum() implementation that works nicely on most platforms for the case where ivsize == nvsize.
And it works correctly (albeit in an ugly and less than optimal way) on most platforms for the case ivsize != nvsize. (I'd like to improve on this aspect.)
But the current uniqnum() implementation is broken on the doubledouble platform. (I'd like to improve on this aspect, too.)
And it's also broken on Windows perls (whose nvsize == ivsize) that have been built with older MS compilers. (I'd like to improve on this aspect, too - though not quite so committed to doing so.)
I have a fix for all of these issues. (My fix for MS compilers needs some refinement as I haven't yet determined the version number beyond which the fix is no longer needed.)

However, in order to achieve these aims, I need to know what, if any, are the perceived problems with this PR.
(And if there are no perceived problems, then please merge.)

Cheers,
Rob

leonerd:
@sisyphus hi. Sorry for the silence on this. I've been away on some travels the past few days. I'll see if I get some time this evening to take a look but if not it'll be a week or two more yet.

leonerd:
Hmm, overall seems fine (or at least, no worse than the existing code) but it is missing the

    sv_catpvn(keysv, "f", 1);

defence code that the other branch has. Please add.

I'm still slightly dubious on the idea of using raw bytes of the underlying representation to distinguish unique values, simply on the grounds that there might be false negatives here; holes in the byte structure on some platform where the representation isn't fully dense, such that two representations can have different bytes even though they represent the same value. As yet I'm not aware of such a platform, but I still feel uneasy at the problem.

But as I said; your code is no worse than what already stands before, so that alone shouldn't prevent it from being applied. Something we can look at afterwards I feel.

sisyphus:
I believe that we don't need that additional sv_catpvn() call in that branch.
Unlike the \#else branch, the \#if NVSIZE > IVSIZE branch doesn't stringify the UV/IV values - it only looks at the bytes of SvNV(arg).
Hence unique strings are guaranteed to relate to unique values, and identical strings are guaranteed to relate to identical values (except, of course, for zeroes and NaNs, which are handled separately).

This PR has no effect on perls whose nvsize and ivsize are both 8.
According to my benchmarking, it measurably improves performance for all other perls by avoiding that stringification of IV/UV values.
( This approach also demonstrates just how simple this exercise would have been if we didn't have to concern ourselves with the perl configuration that has ivsize and nvsize both set to 8.)

I put that benchmark script in my initial post to this thread.
It's not very readable, so I'll attach that script to this post and hopefully that will make it more easily understandable if anyone wants to examine it.
It just compares the cost of stringifying UV/IVs (as the \#else branch does) to the cost of handling UV/IVs as NVs (as the \#if NVSIZE > IVSIZE branch does).

Incidentally, I don't think that sv_catpvn() call is needed in the \#else branch, either. It certainly *would* be needed if we were reading the bytes of *all* NVs.
However, we only read the bytes of a certain range of values which, I think, makes it impossible for the clash (that's avoided by the sv_catpvn call) to ever arise in the first place.
But this belief of mine is as yet unproven .... and can certainly wait for another day.
(If someone can come up with an example that proves the need for that sv_catpvn call, then that would definitely save me a lot of head scratching ;-)

Cheers,
Rob
[uniq_bench.TXT](https://github.com/Dual-Life/Scalar-List-Utils/files/4211298/uniq_bench.TXT)

sisyphus:
Ooops ... just realized I didn't address one of @leonerd's concerns in my post (above) of a few minutes ago.
That concern he expressed was:

--quote--
I'm still slightly dubious on the idea of using raw bytes of the underlying representation to distinguish unique values, simply on the grounds that there might be false negatives here; holes in the byte structure on some platform where the representation isn't fully dense, such that two representations can have different bytes even though they represent the same value.
--end quote--

That would be a drastically flawed system, and I don't think it exists -and even if it does exist then the reliability of stringification of values is also seriously questioned.

If 2 (non-zero, non-NaN) NVs have different byte structures then they have different values.
If they have the same byte-structure then they have the same value.
That's true by definition.

The only concern could be that uniqnum reads in 2 different byte structures as being identical, or 2 identical byte structures as being different - and heaven help the poor bugger that has such a fundamentally flawed system.
Sure - it might be that a user assigns 2 values that ought to assign to different NV values, only to have them assigned to the same NV value - something which was not uncommon pre-perl-5.30.0.
But concern about that is not part of uniqnum's charter - uniqnum's charter is to adjudicate on the values it has been given, not on the values the user thought he was providing.
And I think it's never less reliable (and always quicker) to base that adjudication on byte structure than it is to base it upon stringification of values.

I guess it would be nice if uniqnum also checked that the values provided to it were as intended by the user .... but I don't think we want to go there ;-)

leonerd:
> However, we only read the bytes of a certain range of values which, I think, makes it impossible for the clash (that's avoided by the sv_catpvn call) to ever arise in the first place.

Presuming IEEE 754 doubles, and ASCII sure; but those are not guaranteed assumptions. There could exist valid floating-point and string encoding systems where such a clash can happen.

> If 2 (non-zero, non-NaN) NVs have different byte structures then they have different values.

Again, only presuming IEEE 754 and no holes in the byte representation. My concern is - to state yet again - with some hypothetical floating-point system which does not use every single bit of its in-memory representation. That there could exist two different bit-patterns in memory, who only differ in the unused "holes" that aren't used for meaningful fields. These two bit-patterns will be considered distinct by this code, but yet numerically contain the same NV because they are identical in all the meaningful bits.

However, as explained: That problem is distinct from this PR, as it already existed beforehand. So that can await another day.

sisyphus:
Yes - I see your point.
It's a pity you raised it - somewhere there's probably a developer working on some half-arsed system who's going to read that and think "Now *that* is an idea I can utilise ..." ;-)

Your point is pretty much the reason that I created #99 

Cheers,
Rob

PR #97, opened 2020-02-08
ListUtil.xs: silence signed/unsigned build warning in List::Util::sample()

demerphq:
"items" from dXSVARS is an I32. Defining count to be 'UV' means
we get warnings about signed/unsigned comparisons, further, reti
is defined as "int" which /could/ be as little as 16 bits, and also
throws signed/unsigned warnings when compared to count.

This patch defines both count and reti to be IV's.

Eg:
    ListUtil.xs: In function ‘XS_List__Util_sample’:
    ListUtil.xs:1263:14: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
	 if(count > items)
		  ^
    ListUtil.xs:1272:16: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
	 while(reti < count) {
		    ^

### DIFF
--- a/ListUtil.xs
+++ b/ListUtil.xs
@@ -22,6 +22,7 @@
 #include <EXTERN.h>
 #include <perl.h>
 #include <XSUB.h>
+#include <math.h>
 
 #ifdef USE_PPPORT_H
 #  define NEED_sv_2pv_flags 1
--- a/ListUtil.xs
+++ b/ListUtil.xs
@@ -1261,8 +1261,8 @@ sample(...)
 PROTOTYPE: $@
 CODE:
 {
-    UV count = items ? SvUV(ST(0)) : 0;
-    int reti = 0;
+    IV count = items ? SvUV(ST(0)) : 0;
+    IV reti = 0;
     SV *randsv = get_sv("List::Util::RAND", 0);
     CV * const randcv = randsv && SvROK(randsv) && SvTYPE(SvRV(randsv)) == SVt_PVCV ?
         (CV *)SvRV(randsv) : NULL;


demerphq:
This also includes a fix for Win32 builds which need to 

```
#include <math.h>
```

to use trunc(). I had to revert the push of 1.54 into core because of this.

xenu:
> to use trunc().

`trunc` is a C99 function, we shouldn't be using it.

demerphq:
On Sat, 8 Feb 2020 at 14:11, xenu <notifications@github.com> wrote:

> to use trunc().
>
> trunc is a C99 function, we shouldn't be using it.
>
OH! I didnt notice that in the output from `man trunc` until now:

SYNOPSIS
       #include <math.h>

       double trunc(double x);
       float truncf(float x);
       long double truncl(long double x);

       Link with -lm.

   Feature Test Macro Requirements for glibc (see feature_test_macros(7)):

       trunc(), truncf(), truncl():
           _ISOC99_SOURCE || _POSIX_C_SOURCE >= 200112L


yves

xenu:
I made a PR (#98) that removes trunc().

leonerd:
Now I've merged #98, do we still need the `<math.h>` here?

demerphq:
On Tue, 11 Feb 2020, 12:17 Paul Evans, <notifications@github.com> wrote:

> Now I've merged #98
> <https://github.com/Dual-Life/Scalar-List-Utils/pull/98>, do we still
> need the <math.h> here?
>

Nope, sorry, I just have yanked that  ommit after xenu's PR but  i forgot.
But the other patch for the type changes are needed to keep win32 builds
warning free please.

Thanks!
Yves

>

leonerd:
Merged by 71b0e0e

PR #99, opened 2020-02-17
In uniqnum: Avoid reading of unused bytes

sisyphus:
First up, a small fix to something I should have noticed earlier:
Change line 1395 of current github version of ListUtil.xs from:

to
sv_setpvn(keysv, (char *) &nv_arg, 8);

This is because, at that point, sizeof(NV) is always 8.

The remaining changes address my concern about reading in all 12 (or 16, as the case might be) of the 80-bit extended precision long double when only 10 are needed.
Firstly, why go to the trouble of reading 12/16 bytes when only 10 are relevant.
Secondly, I have concerns that the bytes beyond the 10th might contain random garbage that would lead to 2 equivalent long doubles being deemed to have unequal values.

### DIFF
--- a/ListUtil.xs
+++ b/ListUtil.xs
@@ -30,6 +30,15 @@
 #  include "ppport.h"
 #endif
 
+/* For uniqnum, define ACTUAL_NVSIZE to be the number *
+ * of bytes that are actually used to store the NV    */
+
+#if defined(USE_LONG_DOUBLE) && LDBL_MANT_DIG == 64
+#define ACTUAL_NVSIZE 10
+#else
+#define ACTUAL_NVSIZE NVSIZE
+#endif
+
 #ifndef PERL_VERSION_DECIMAL
 #  define PERL_VERSION_DECIMAL(r,v,s) (r*1000000 + v*1000 + s)
 #endif
@@ -1364,8 +1373,12 @@ CODE:
             sv_setpvf(keysv, "%" NVgf, nv_arg);
         }
         else {
-            /* Use the byte structure of the NV. */
-            sv_setpvn(keysv, (char *) &nv_arg, sizeof(NV));  
+            /* Use the byte structure of the NV.            *
+             * ACTUAL_NVSIZE is 10 if NV type is the        *
+             * 10-byte extended precision long double.      *
+             * Else ACTUAL_NVSIZE is the same as sizeof(NV) */
+
+            sv_setpvn(keysv, (char *) &nv_arg, ACTUAL_NVSIZE);  
         }
 #else                 /* close nvsize>ivsize block, open ivsize==nvsize block */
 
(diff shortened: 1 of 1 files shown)


sisyphus:
BTW, I don't have any additional PRs to make in the immediate future.

It'd be great if we could get this PR applied and a new Scalar-List-Utils release made so that we can subject it to wider testing and fix any issues in time for perl-5.32.0.

Cheers,
Rob

sisyphus:
I've come up with a much cleaner (and slightly faster) way of dealing with the case where IVSIZE == NVSIZE == 8.
(I hadn't expected to come up with such an approach so soon, but it turned out to be easier than I had anticipated. So I'll outline it here, but I won't submit a PR for it until after this one has been dealt with.)
Now would be a good time for comments upon it, and especially for any ideas that would improve it.

Basically, it enables us to do away with the Windows-specific stuff  - there's no longer a need to detect mingw compilers, nor do we need to define __USE_MINGW_ANSI_STDIO, and it also works with *all* microsoft compilers - not just the recent ones.
Anyway ... if there's anyone interested, take a look at the attached new_stuff.txt

[new_stuff.txt](https://github.com/Dual-Life/Scalar-List-Utils/files/4249864/new_stuff.txt)

sisyphus:
Tonight, David (who else) Mitchell presented a simpler method of determining whether a UV can be represented exactly as an NV at:
https://www.perlmonks.org/?node_id=11113426

I'll go with his suggestion when I make that next PR.

Cheers,
Rob

sisyphus:
It eventually turned out that Dave Mitchell's idea for a solution is susceptible to the very problem I was trying to work around.
Anyway ... enough with speculation about what I want to change next ... could we just get this PR applied ... or get some feedback on what's unacceptable about it.

I'm starting to worry about the appalling possibility of perl-5.32.0 core shipping with either S-L-U-1.53 or S-L-U 1.54.

Cheers,
Rob

leonerd:
Seems a bit fragile, but probably no more fragile than it was originally.

sisyphus:
In terms of how it's testing (List-Uniqnum-0.05) it seems quite robust:
453 ALLs 452 PASSes 1 UNKNOWN.

The one "UNKNOWN" makes no sense to me. I doubt very much that it has anything to do with the actual XS code.

The only possible point of fragility I can see is its reliance on the C toolchain to correctly cast an IV/UV to an NV.
And, remember, such casts are being performed *only* upon IV/UVs that can be *exactly* represented as an NV.
IIRC, all other IV/UVs are being stringified with this particular code.
If the C toolchain can't correctly cast such an IV/UV to an NV then I think there are bigger concerns than just Lit::Util::uniqnum.

Anyway, I'd better investigate the reason that perl-5.6.2 is choking on #100

PR #100, opened 2020-03-20
Fix uniqnum for less recent Microsoft Compilers

sisyphus:
Also fixes a problem I came across with DoubleDouble NV types.
The problem was, that with this NV type, the DoubleDouble pair (x, 0.0) would be considered distinct from the DoubleDouble pair (x, -0.0).
Clearly, they should be regarded as duplicates of each other because x + 0.0 == x - 0.0 for all non-NaN x.

The main fix, however, is for perls whose nvsize == ivsize == 8 that were built with older MS compilers like VC++ 2010 (aka Visual C++ 10.0, among other things).
This was the issue that for many large IV/UV values that could also be exactly represented by an NV,
the NV and the IV/UV were not being recognized as duplicates of each other.

To fix it, I was forced to a solution that turned out to be so elegant and efficient, that I applied it "across the board" for all perls whose nvsize == ivsize == 8.

I'm not sure if this PR forum allows one to see what the uniqnum function in it's proposed form will actually look like.
I'm guessing it does, but if not, feel free to take a look at the uniqnum() implementation in https://fastapi.metacpan.org/source/SISYPHUS/List-Uniqnum-0.08/Uniqnum.xs as it should be identical to the uniqnum implementation proposed here - save for some small differences in comments.
(That implementation of Uniqnum.xs is passing all tests. There is just the one UNKNOWN result which presumably relates to the brokenness of that particular Strawberry Perl smoker, as I strike no such problem with the exact same build of Strawberry Perl.)

That's about it ... comments and questions welcome ... now it's time to find out what Travis thinks about this ...

### DIFF
--- a/ListUtil.xs
+++ b/ListUtil.xs
@@ -3,21 +3,6 @@
  * modify it under the same terms as Perl itself.
  */
 
-/* For MinGW-built perls, define __USE_MINGW_ANSI_STDIO unless  *
- * perl was built with that symbol defined. We need this for    *
- * "%.20g" formatting - and it must be done done prior to the   *
- * inclusion of all headers.                                    */
-
-#if defined(__MINGW32__) && !defined(__USE_MINGW_ANSIO_STDIO)
-
-#  define __USE_MINGW_ANSI_STDIO 1
-
-   /* Identify that perl itself was built  *
-    * without -D__USE_MINGW_STDIO_ANSI     */
-#  define MINGW_PERL_NO_ANSI 1
-
-#endif
-
 #define PERL_NO_GET_CONTEXT /* we want efficiency */
 #include <EXTERN.h>
 #include <perl.h>
@@ -34,10 +19,16 @@
  * of bytes that are actually used to store the NV    */
 
 #if defined(USE_LONG_DOUBLE) && LDBL_MANT_DIG == 64
-#define ACTUAL_NVSIZE 10
+#  define ACTUAL_NVSIZE 10
 #else
-#define ACTUAL_NVSIZE NVSIZE
+#  define ACTUAL_NVSIZE NVSIZE
 #endif
+
+/* Detect "DoubleDouble" nvtype */
+
+#if defined(USE_LONG_DOUBLE) && LDBL_MANT_DIG == 106
+#  define NV_IS_DOUBLEDOUBLE
+#endif  
 
 #ifndef PERL_VERSION_DECIMAL
 #  define PERL_VERSION_DECIMAL(r,v,s) (r*1000000 + v*1000 + s)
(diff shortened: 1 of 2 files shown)


sisyphus:
Not sure why perl-5.6.2 is failing those test just yet - though it looks like it's the same issue for all of the failures.
It's a pity I can't build x86_64-Linux perl-5.6.2 on Ubuntu.
Anyone here managed to do that ?

In the meantime, I'll just have to upload one or more debug versions of uniq.t (and perhaps other files) and see if I can work out what's going on from the ensuing Travis results.

sisyphus:
The problem was that uniqnum() was altering the numeric flags of its arguments - in a way that made perl-5.6.2 stringify things in an unexpected way.
Of course, it's probably better that uniqnum() doesn't alter the flags of its arguments.
And this is now the case - as evidenced by the fact that perl-5.6.2 is now passing the test suite.

This is achieved by doing either:
nv_arg = (NV)SvUV(arg);
or
nv_arg = (NV)SvIV(arg);
(whichever is appropriate), instead of just doing the cover-all-bases:
nv_arg = SvNV(arg);
(which was being done in a couple of places).

I've uploaded List-Uniqnum-0.09, which mirrors this change, so we can check that it tests ok with the various testers.

Is this PR still in an acceptable form to understand ... or is a new PR that avoids all of the intermediate changes preferable ?

Cheers,
Rob

sisyphus:
BTW, I don't have any more changes to make at the moment.
I keep thinking that there's a better way to handle the issue with DoubleDoubles, but I haven't found it yet, and it can wait if it even exists.

IMO, with the application of this PR, uniqnum() would finally be fit for purpose.

Cheers,
Rob

ListUtil.xs
```diff
@@ -1357,75 +1348,98 @@ CODE:
...
+             * never used. For all other NV types ACTUAL_NVSIZE == sizeof(NV). */
             sv_setpvn(keysv, (char *) &nv_arg, ACTUAL_NVSIZE);  
         }
-#else                 /* close nvsize>ivsize block, open ivsize==nvsize block */
+#else                                    /* $Config{nvsize} == $Config{ivsize} == 8 */ 
+        if( SvIOK(arg) || !SvOK(arg) ) {
 
-            /* Because IVSIZE == NVSIZE we cannot rely *
-             * solely on the value of SvNV(arg)        */
+           /* It doesn't matter if SvUOK(arg) is TRUE */
```
leonerd: comment is indented oddly

ListUtil.xs
```diff
@@ -1357,75 +1348,98 @@ CODE:
...
+            IV iv = SvIV(arg);
+
+           /* use "0" for all zeros */
```
leonerd: ditto

ListUtil.xs
```diff
@@ -1357,75 +1348,98 @@ CODE:
...
-                    char buffer[32];
-                    sprintf(buffer, "%0.20" NVgf, nv_arg);
-                    sv_setpvf(keysv, "%s", buffer);
-#else
-                    sv_setpvf(keysv, "%0.20" NVgf, nv_arg);
-#endif
+                    /* Read in the bytes, rather than the numeric value of the IV/UV as  *
+                     * this is more efficient, despite having to sv_catpvn an extra byte.*/
+                    sv_setpvn(keysv, (char *) &iv, 8);
+                    /* We add an extra byte to distinguish between an IV/UV and an NV.   *
```
leonerd: Does this block suffer the problem that it considers all IVs distinct from all UVs? Does that mean it's possible to represent the same (positive) value as both an IV and a UV and yet they compare distinctly here?

sisyphus:
> Does this block suffer the problem that it considers all IVs distinct from all UVs?

No.
By the time we've got to this part of the code, it's already established that we have an IV.
All this block is does is to separate the IVs for which IsUV is set (giving them the "U" suffix) from the IVs for which IsUV is unset (giving them the"I" suffix).

There is no intersection wrt *values* in the 2 groups.
IsUV being set implies a value of greater than or equal to 1 << 63.
IsUV being unset implies a value of less than  1 << 63.

In terms of bit structure, there *is* of course intersection between the 2 groups ... wherein lies the need for the different suffix.

IsUV can only ever be set if the 64th bit is set - in which case the value is, by definition, unsigned.
If the 64th bit is set and IsUV is unset, then the value is, by definition, signed (and negative).

You won't ever get a case where some value x sometimes has the IsUV flag set, and other times unset.

Cheers,
Rob

sisyphus:
Although I forgot to specifically mention it at the time, the misaligned comments were fixed 4 days ago.
I'm just pointing it out now in case there's a delay caused by some misconception that it hasn't been done yet.

Cheers,
Rob

ListUtil.xs
```diff
@@ -1357,75 +1348,98 @@ CODE:
...
-            sv_setpvs(keysv, "0");
-        }
-        else if (nv_arg != nv_arg) {
-            /* for NaN, use the platform's normal stringification */
-            sv_setpvf(keysv, "%" NVgf, nv_arg);
+#if NVSIZE > IVSIZE                          /* $Config{nvsize} > $Config{ivsize} */
+        /* Avoid altering arg's flags */ 
+        if(SvUOK(arg))      nv_arg = (NV)SvUV(arg);
+        else if(SvIOK(arg)) nv_arg = (NV)SvIV(arg);
+        else                nv_arg = SvNV(arg);
```
leonerd, line 1355: This part slightly worries me. Won't it get confused by an IV+NV dualvar, such as those created by:

```perl
# uniqnum not confused by IV'ified floats
{
  my @nums = ( 2.1, 2.2, 2.3 ); # all 2 as IVs
  map { sprintf "%d", $_ } @nums;
  # All @nums now have both NOK and IOK
  is( scalar( uniqnum @nums ), 3, 'uniqnum not confused by dual IV+NV' );
}
```

I'm currently building myself a `-Duselongdouble` perl so I can test on a perl where `NVSIZE > IVSIZE` to see for myself...

leonerd: Update: Seems like this is fine. The SVs gain the private pIOK flag but not the (public) IOK one, so the `uniqnum` code isn't confused:

```
SV = PVNV(0x5600c853bbd0) at 0x5600c8c59ba0
  REFCNT = 2
  FLAGS = (NOK,pIOK,pNOK)
  IV = 2
  NV = 2.1
  PV = 0
SV = PVNV(0x5600c853bba0) at 0x5600c8c5a438
  REFCNT = 2
  FLAGS = (NOK,pIOK,pNOK)
  IV = 2
  NV = 2.2
  PV = 0
SV = PVNV(0x5600c853bb70) at 0x5600c8c59cd8
  REFCNT = 2
  FLAGS = (NOK,pIOK,pNOK)
  IV = 2
  NV = 2.3
  PV = 0
ok 33 - uniqnum not confused by dual IV+NV
```

leonerd:
I think I'm going to merge this as it is, but then have quite a tidy-up of the actual implementation code. It's getting very messy now. The original structure of using a common `ALIAS`ed block of code for the three `uniq*()` functions is now becoming too cluttered. I'll probably move `uniqnum` to its own copy to remove the `ix` tests. While I'm there I might see if I can rework the logic around to pull out common structures a bit better too. Hopefully there's unit-tests to cover all the bases. :)

ListUtil.xs
leonerd: Ah turns out on Perl 5.6 that doesn't hold. It fails:

https://travis-ci.org/github/Dual-Life/Scalar-List-Utils/jobs/672183461

sisyphus: The same test fails in the same way with List-Util-1.53.
This looks to me like a bug in the way that perl-5.6.2 is handling dualvars.
I would think dwimmery dictates that
```
map { sprintf "%d", $_ } @nums;
```
should still leave the elements of \@nums being regarded primarily as NVs.
But, if all 4 of IOK , pIOK, NOK and pNOK are set (as is the case with 5.6.2), then who is to say which value to consider ?

Interestingly, on 5.6.2, print() still outputs the NV values of @nums.
I guess that print() must be searching for NOK before it goes for looking for IOK.
Whereas uniqnum() has always looked for IOK first.

I think we should skip that test for any perl that sets dualvars so ambiguously.

Cheers,
Rob

PR #102, opened 2020-04-08
AMG_CALLun() is poor style; use AMG_CALLunary() instead

leonerd:


leonerd:
On second thoughts, `AMG_CALLunary()` doesn't exist before 5.14 and we're still supposed to be supporting back to 5.6. So this will have to stay :(

PR #103, opened 2020-04-08
Add a `uniqint()` function

leonerd:
To smooth over most of the complications of `uniqnum()` when you know your numbers should be well-behaved integers.

### DIFF
--- a/ListUtil.xs
+++ b/ListUtil.xs
@@ -1306,6 +1306,7 @@ void
 uniq(...)
 PROTOTYPE: @
 ALIAS:
+    uniqint = 0
     uniqstr = 1
     uniq    = 2
 CODE:
@@ -1346,6 +1347,12 @@ CODE:
                 ST(retcount) = arg;
             retcount++;
             continue;
+        }
+        if(ix == 0) {
+            /* uniqint */
+            /* coerce to integer */
+            arg = sv_2mortal(SvUOK(arg) ? newSVuv(SvUV(arg)) :
+                                          newSViv(SvIV(arg)));
         }
 #ifdef HV_FETCH_EMPTY_HE
         he = (HE*) hv_common(seen, arg, NULL, 0, 0, HV_FETCH_LVALUE | HV_FETCH_EMPTY_HE, NULL, 0);
--- a/ListUtil.xs
+++ b/ListUtil.xs
@@ -1351,7 +1351,10 @@ CODE:
         if(ix == 0) {
             /* uniqint */
             /* coerce to integer */
+#if PERL_VERSION >= 8
+            /* int_amg only appeared in perl 5.8.0 */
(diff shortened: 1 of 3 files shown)


sisyphus:
On a perl whose ivsize is 8:
```
C:\>perl -Mblib -MList::Util="uniqint" -wle "@x = uniqint(1 << 63, 2 ** 63); print for @x;"
9223372036854775808
-9223372036854775808
```
On a perl whose ivsize is 4:
```
C:\>perl -Mblib -MList::Util="uniqint" -le "@x = uniqint(1 << 31, 2 ** 31); print for @x;"
2147483648
-2147483648
```
This might be something that's afflicting only those 2 values - ie the smallest integer value that sets the IsUV flag.
(Haven't had time to check thoroughly.)

Cheers,
Rob

sisyphus:
Also, uniqint is slower than uniqnum.
Here's the benchmark script I used:
```
use warnings;
use List::Util qw(uniqint uniqnum);
use Benchmark;
use Test::More tests => 3;

$v = 1000000;

our @in = (-$v .. $v);

timethese(1, {

'a_uniqnum' => '$c2 = uniqnum(@in);',
'b_uniqint' => '$c1 = uniqint(@in);',
'c_uniqnum' => '@c2 = uniqnum(@in);',
'd_uniqint' => '@c1 = uniqint(@in);',
});

is($c2, 2000001);
is($c1, 2000001);
is_deeply([@c2], [@c1], "arrays match");
```
And here's the output I got:
```
C:\_32\pscrpt\scalar-list-utils>perl try.pl
1..3
Benchmark: timing 1 iterations of a_uniqnum, b_uniqint, c_uniqnum, d_uniqint...
 a_uniqnum:  2 wallclock secs ( 1.50 usr +  0.06 sys =  1.56 CPU) @  0.64/s (n=1)
 b_uniqint:  2 wallclock secs ( 1.81 usr +  0.08 sys =  1.89 CPU) @  0.53/s (n=1)
 c_uniqnum:  1 wallclock secs ( 1.40 usr +  0.06 sys =  1.47 CPU) @  0.68/s (n=1)
 d_uniqint:  2 wallclock secs ( 1.70 usr +  0.08 sys =  1.78 CPU) @  0.56/s (n=1)
ok 1
ok 2
ok 3 - arrays match
```
Cheers,
Rob

leonerd:
> This might be something that's afflicting only those 2 values - ie the smallest integer value that sets the IsUV flag.
> (Haven't had time to check thoroughly.)

Ah yes; now fixed in the two commits 133f42b, cdb1b3c

As to performance, these changes also avoid making a mortal copy of the SVs in most cases, so it should be a bit faster as well. In fact now the performance of the two functions is almost identical. `uniqint()` still slightly slower, but that comes with the bonus that it squashes its values to integers, and can cope with overloaded maths values like bigint; which would confuse the more generic `uniqnum()`.

PR #105, opened 2020-04-17
Fixes to uniqint

sisyphus:
The uniqint() function was not correctly handling the doubles 2**64 and NaN.
These proposed changes to ListUtil.xs will fix those 2 cases, while (AFAIK) not breaking anything else.

Also proposed are changes to t/uniq.t, which will verify that these (and other previously untested) aspects of uniqint are, indeed, working as intended.

Cheers,
Rob


sisyphus:
Is uniqint actually of any significant use ?
AFAICT it provides nothing that was not already provided by uniqstr.

Additionally, uniqstr does not suffer the bugs that this PR addresses.

Cheers,
Rob

Grinnz:
Of course it does, numeric equality and string equality are very different. 0.0 and 0 and 0e0 are all the same integer. And apart from the float complexities uniqnum has to deal with, this serves the much more common and simple use case of integer uniqueness.

sisyphus:
I was looking at the sorts of arguments that one might normally pass to uniqint - namely IVs, UVs, Math::BigInt/Math::GMP objects, NVs that contain integer values.

AFAICT, for these arguments, uniqint and uniqstr behave identically.

But, you're right - if you insert some PVs into the argument list, then uniqint and uniqstr can behave differently, because uniqint looks at the PV in numeric context whereas it's just a string to uniqstr.

Also, if the argument list contains NVs holding fractional values, then the behaviours can differ.

So, I can see that uniqint has the potential to be handy for some cases.

Actually, I've only just realized that the input array (1 << $shift, 2 ** $shift) can return differently for uniqint and uniqstr -  depending upon the values of ivsize, nvsize and $shift.
(I should, at least, have picked up on that earlier.)

Cheers,
Rob

t/uniq.t
```diff
@@ -83,6 +83,85 @@ is_deeply( [ uniqint 6.1, 6.2, 6.3 ],
...
+
+my @correct = (1 << $ls, 1 << ($ls - 3), 5 << ($ls -3));
+
+if( $Config{ivsize} == 8 && $Config{nvsize} == 8 ) {
+
+     # Add some more UV-NV pairs of equivalent values.
+     # Each of these values is exactly representable
+     # as either a UV or an NV.
+
+     push @in, ( 9007199254740991,     9.007199254740991e+15,
```
leonerd, line 108: This chunk has a lot of "magic numbers" in it with no suggestion of how to recreate those. It would be great if they could be generated more programmatically somehow, similar to the `@correct` case above so it's more apparent where they come from.

leonerd:
I'm happy with the `NaN` and `2 ** $ls` cases; those can go in as they stand. But I'm concerned about the long list of magical numbers. Can you either rewrite that in a way which makes it more obvious where they came from, or at least split this PR into two parts so I can merge the former while we continue to discuss the latter?
Thanks,

sisyphus:
Added some explanatory comments, and showed how each number is derived.
I don't mind if those tests are removed. (Just ping me and I'll do it.)
They include values that were problematic with uniqnum.
They haven't been an issue with uniqint, but I think it doesn't hurt to check them, especially since there have been issues with a couple of numbers of a similar type (2 ** 63 and 2 ** 64).

I think they also serve as a guard against future bug intrusion.

Cheers,
Rob

sisyphus:
Test 16 ( "uniqint correctly compares UV/IVs that don't overflow NVs" ) was failing to catch a failure wrt the two values 1.8446744073709549568e+19 and 1.8446744073709139968e+19 on perls built with Microsoft's VC 2010.

The same two values also led to a (different) failure with VC 7.1, though this particular failure was caught.

Anyway, I therefore removed those 2 values (and their equivalent UV forms) from test 16, and created another test ( "uniqint correctly compares UVs that don't overflow NVs" ) which detects both types of failure.

With VC 10.0, the failure is:
```
t\uniq.t .............. 1/35
#   Failed test 'uniqint handles NVs that equate to UVs'
#   at t\uniq.t line 160.
#     Structures begin differing at:
#          $got->[0] = '9223372036854775808'
#     $expected->[0] = '13835058055282163712'
# Looks like you failed 1 test of 35.
```
For some reason, with that compiler, doubles greater than 2 ** 63 and less than 2 ** 64 all seem to end up as the UV 9223372036854775808
With VC 7.1, the failure is:
```
t\uniq.t .............. 1/35
#   Failed test 'uniqint correctly compares UVs that don't overflow NVs'
#   at t\uniq.t line 148.
#     Structures begin differing at:
#          $got->[0] = '13835058055282163711'
#     $expected->[0] = '13835058055282163712'
# Looks like you failed 1 test of 35.
```
For some reason, with that compiler, doubles greater than 2 ** 63 and less than 2 ** 64 typically end up as a UV whose value is one less than it ought to be.

With VC 14.2, all tests pass.

At the moment, I don't properly understand the rules that are leading to these failures,
Nor do I know what is needed to fix them.

Cheers,
Rob

sisyphus:
Hopefully test 17 of t/uniq.t is now fixed for perl-5.6.2

sisyphus:
I have a fix for the problem with those troublesome Microsoft toolchains that cannot handle the casting of many NVs to the correct UV values.
Here's a diff that contains that proposed fix.
Do you want it added to this PR ?

It hinges on the value of the compiler symbol _MSC_VER, and I've chosen that value to be "1900", though I'm not certain that's the correct value.
If "1900" is not high enough, we may well find out one day when someone gets a test failure.
If the value could be set lower than 1900, then we may never find out, as everything will still work fine even if the workaround is unnecessary.
Here's the proposed change:
```
--- ListUtil.xs_orig	2020-05-01 00:36:05 +1000
+++ ListUtil.xs	2020-05-01 10:17:35 +1000
@@ -1375,9 +1375,24 @@
                 else if(nv <= (NV)IV_MIN)
                     /* Too negative for IV - use NV */
                     arg = newSVnv(Perl_ceil(nv));
-                else if(nv > 0 && (UV)nv > (UV)IV_MAX)
+                else if(nv > 0 && (UV)nv > (UV)IV_MAX) {
                     /* Too positive for IV - use UV */
+#if defined(_MSC_VER) && _MSC_VER < 1900 && IVSIZE == 8 && NVSIZE == 8
+                    if(nv > 9223372036854775808.0 && nv < 1.8446744073709552e+19 )
+
+                        /*****************************************************
+                         * Less recent versions of Visual Studio make a mess *
+                         * of casting NVs that are greater than 2 ** 63 and  *
+                         * less than 2 ** 64 to the correct UV.              *
+                         * Hence, we use this workaround.                    *
+                         * Is "1900" the correct _MSC_VER value to specify ? *
+                         *****************************************************/
+                   
+                        arg = newSVuv( 9223372036854775808 + (UV)(nv - 9223372036854775808.0) );
+                    else
+#endif 
                     arg = newSVuv(nv);
+                }
                 else
                     /* Must now fit into IV */
                     arg = newSViv(nv);
```

But there's another more wide reaching problem, that I should have thought of earlier, because it's similar to the problem we faced with uniqnum - where perl's braindead print() implementation provides strings of insufficient precision to allow the distinguishing of many different NV values from each other.
This affects all of the perl ivsize/nvsize configurations. Here's a demo of a couple of values for a perl whose nvsize is 8:
```
use warnings;
use Config;
use Test::More tests => 3;
use List::Util qw(uniqint);

die "Run this demo only on perls whose nvsize is 8"
  unless $Config{nvsize} == 8;

$x1 = 3.689348814741907e19;
print scalar reverse unpack "h*", pack "d<", $x1;
print "\n";

$x2 = 3.6893488147419099e19;
print scalar reverse unpack "h*", pack "d<", $x2;
print "\n";

cmp_ok($x1, '!=', $x2, '$x1 != $x2') ;
cmp_ok(int $x1, '!=', int $x2, 'int $x1 != int $x2') ;

$c = uniqint( $x1, $x2);

cmp_ok($c, '==', 2, 'uniqint regards $x1 and $x2 as unique');
```
Which outputs:
```
1..3
43fffffffffffff8
43ffffffffffffff
ok 1 - $x1 != $x2
ok 2 - int $x1 != int $x2
not ok 3 - uniqint regards $x1 and $x2 as unique
#   Failed test 'uniqint regards $x1 and $x2 as unique'
#   at try.txt line 25.
#          got: 1
#     expected: 2
# Looks like you failed 1 test of 3.
```
One possible solution is to ban NVs greater than ~0 from uniqint's argument list - and have it croak if any such values are encountered.

Otherwise, instead of relying on print(), we use printf() with a suitable precision for the particular nvtype - which is essentially what we tried for a while with uniqnum().

I think that will work with uniqint - and I'll proceed to work on that as time permits.
It should have minimal effect on the performance of uniqint.

In the meantime, please let me know if you want that _MSC_VER fix pushed to my fork of the Scalar-List-Utils repo.

Cheers,
Rob

sisyphus:
I still struggle to see the point of uniqint.
Sometimes I get a glimpse of how it might be useful, but I've never quite managed to nail it.
And I definitely don't understand the current spec.

Let me illustrate the conundrum I'm currently facing.
Let's say we have (on a perl whose nvsize is 8) :
```
$nv1 = 3.689348814741907e19;  # 0x1.ffffffffffff8p+64
$nv2 = 3.6893488147419099e19; # 0x1.fffffffffffffp+64
$count = uniqint($nv1, $nv2);
```
Note that $nv1 and $nv2 represent different integer values, both of which are greater than UV_MAX.

The one and only question I have (and would dearly love to be answered) is:

By what mechanism is it envisaged that $count will be set to the correct value of 2 ?

Currently $count will be set to 1 - because, perl stringifies $nv1 and $nv2 to the same string of '3.68934881474191e+19'.
(And note that $h{$nv1} and $h{$nv2} would be one and the same hash key, despite that $nv1 != $nv2 .... which, admittedly, is phenomenally braindead behaviour .... but it's what we're presently stuck with.)

OTOH, uniqnum gets it right - and I guess we could always resort to the same technigue used there.
However, I get the feeling that there's a desire to achieve the result by some means other than that used by uniqnum .... but I'm not sure what that "other" means is .... hence the "one and only question" that I've asked.

I'm thinking that instead of doing (as is currently happening for these values):
arg = newSVnv(Perl_floor(nv));

We should do (when nvsize is 8):
char buffer[32];
sprintf(buffer, "%.17g", Perl_floor(nv));
arg = newSVpv(buffer, 0);

Is that a good enough approach to take ? (Dammit ... now I've asked a second question ...)
One possible issue with that might be that if you do:
```
@u = uniqint($nv1, $nv2);
```
then the array that's returned is populated with 2 PVs - not with NVs, as the current implementation does.
I've no idea whether that would matter.
(You might as well consider that to be a third question.) 

BTW, I note that, if ivsize is also 8, then the Perl_floor() call is unnecessary because nv has to be an integral value, anyway - so there's a small optimization to be made there, too.

In any case, it seems to me that uniqint is nowhere near ready for perl core, and should be removed from List::Util until we've had a proper opportunity to get it sorted out.

Cheers,
Rob

sisyphus:
It's probably not worth applying this PR as I've encountered additional bugs - for which I'll hopefully soon have fixes and tests.
Trying to look at the actual values is a nightmare (and the reason I abandoned it in preference to byte structure) - but I think I might be nearly through it.

Cheers,
Rob

sisyphus:
Duh ... when I said "and the reason I abandoned it in preference to byte structure" I meant "and the reason I abandoned it in preference to byte structure for uniqnum".
I haven't "abandoned it" wrt uniqint (yet).

Cheers,
Rob

sisyphus:
Closing in preparation for next PR.

Cheers,
Rob

PR #108, opened 2020-05-07
Fixes re List::Util::uniqint

sisyphus:
This PR addresses all of the uniqint issues that I've come across to date.
I don't guarantee that there are no more issues yet to be uncovered.

A good sample of the problems presented by the current implementation of uniqint (in Scalar-List-Utils-1.55) can be seen by running this PR's modified t/uniq.t against Scalar-List-Utils-1.55
Also tidied up a test in t/uniqnum.t.

Cheers,
Rob

### DIFF
--- a/ListUtil.xs
+++ b/ListUtil.xs
@@ -15,6 +15,11 @@
 #  include "ppport.h"
 #endif
 
+/* Detect "DoubleDouble" nvtype */
+#if defined(USE_LONG_DOUBLE) && LDBL_MANT_DIG == 106
+#  define NV_IS_DOUBLEDOUBLE
+#endif
+
 /* For uniqnum, define ACTUAL_NVSIZE to be the number *
  * of bytes that are actually used to store the NV    */
 
@@ -24,10 +29,22 @@
 #  define ACTUAL_NVSIZE NVSIZE
 #endif
 
-/* Detect "DoubleDouble" nvtype */
+/* Define some additional symbols for uniqint */
+#if ACTUAL_NVSIZE == 8
+#  define NV_DEC_PREC 17
+#  define NV_BUF_SIZE 32
 
-#if defined(USE_LONG_DOUBLE) && LDBL_MANT_DIG == 106
-#  define NV_IS_DOUBLEDOUBLE
+#elif ACTUAL_NVSIZE == 10
+#  define NV_DEC_PREC 21
+#  define NV_BUF_SIZE 32
+
+#elif defined(NV_IS_DOUBLEDOUBLE)
+#  define NV_DEC_PREC 33
+#  define NV_BUF_SIZE 48
+
+#else
+#  define NV_DEC_PREC 36
+#  define NV_BUF_SIZE 48
 #endif
 
 #ifndef PERL_VERSION_DECIMAL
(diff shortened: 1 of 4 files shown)


zhmylove:
@sisyphus I see a lot of improvements for `uniqint` here.
May I ask you to fix also a SIGSEGV in it?

I'd suggest to add also these tests:
```perl
use List::Util qw( uniqint );
use Math::BigInt;
my $obj;
$obj = Math::BigInt->new(5);
is_deeply( [ uniqint $obj, $obj ],
           [ $obj ],
           'uniqint compares Math::BigInt objects' );

$obj = bless {}, __PACKAGE__;
is_deeply( [ uniqint $obj, $obj ],
           [ $obj ],
           'uniqint compares non-numerical objects' );
```

Segmentation violation exists both in current and your code.

@leonerd am I right that we don't want any segfaults from such a code?

sisyphus:
I think this should be treated as a separate issue, as it's not related to anything being done by this PR (AFAICS).
Could you do that please ?

You can work around it by using <code>$obj</code> in numeric context prior to passing it to uniqint().
Or you can just interpolate <code>$obj</code> it prior to passing it to uniqint().
That allows uniqint() to run and the test to pass.
Here's a demo that works (I guess) as expected:
```
use strict;
use warnings;
use List::Util qw(uniqint);
use Math::BigInt;
use Test::More;

my $obj = bless {}, __PACKAGE__;

$obj < 0; # Allows uniqint() to run.
# "$obj";   # Also allows uniqint() to run.

is_deeply( [ uniqint $obj, $obj ],
           [ $obj ],
           'uniqint compares non-numerical objects' );

done_testing();
```
Of course, it also outputs a "Useless use of ..." warning, even though we have just proven that our "use" was <b>not</b> "useless" ;-)

A proper fix would address the issue inside ListUtil.xs such that the segfault is avoided.
I had a  go at getting that to work, but couldn't quickly find the correct method. I'll have another try when I get the time and motivation.
This PR has sat here untouched for over 4 years - and that doesn't really help much as regards "motivation".

sisyphus:
Some more digging, and inside uniq() in ListUtils.xs, there's the line of code (line 1376):
```
if(!SvOK(arg) || SvNOK(arg) || SvPOK(arg))
```
The segfault is happening when the <code>!SvOK(arg)</code> condition is being tested.

That's odd - because, in another script, I had already checked to see whether <code>SvOK(arg)</code> was true :
```
use warnings;
use strict;
use Inline C =><<'EOC';

int foo(SV * in) {
  if(!SvOK(in)) return 0;
  return 1;
}

EOC

my $obj = bless {}, __PACKAGE__;
print foo($obj);
```
That all ran fine, and produced the output of "1".
So there's apparently a problem with the way that this <code>$obj</code> is being accessed by uniq().

Hopefully there are people here who can pinpoint the problem.
I'll keep pursuing, but I think I might be getting into an area that's outside my comfort zone.

zhmylove:
@sisyphus thanks for your research and help!

Actually, the issue is with `(arg = AMG_CALLun(arg, int))` as `SvAMAGIC(arg)` returns TRUE.
I'll continue the investigation further and maybe create a fix for the problem.

So you think it'd be better to deliver it via another PR?

zhmylove:
Done with it: https://github.com/Dual-Life/Scalar-List-Utils/pull/137

sisyphus:
Looks good !!
https://github.com/Dual-Life/Scalar-List-Utils/pull/137 works well for me.

PR #109, opened 2020-05-28
Update isdual docs to represent its (lack of) usefulness

Grinnz:
Also fix the description of `$!` since isdual actually returns true for it.


leonerd:
LGTM. This is a good thing to point out.

PR #111, opened 2020-08-11
Update ppport.h to version 3.60

atoomic:
Use PERL_VERSION compare macros from ppport.h
Also add 5.32 to the Travis CI matrix.


leonerd:
For unrelated reasons I've now updated it to 3.62. I can close this as a duplicate.

PR #113, opened 2021-01-05
Fix segfaults when undefined subroutine passed to block-taking subs

ilmari:
The pair* functions weren't checking the sv_2cv return value at all,
while the rest were only checking for not a subroutine, not for
references to undefined subs.


leonerd (approved):
LGTM

PR #115, opened 2021-09-06
Correct refcount handling on hv_store_ent to PL_DBsub in set_subname

nwc10:
The code had been handling the return value from hv_store_ent incorrectly.
hv_store_ent returns NULL if the store didn't happen (a tied hash), or if
the store operation failed - in other words, it returns true *if* it took
ownership of a reference to the stored value, false if not. This entire API
is somewhat clunky and error prone, and also somewhat makework, as store
never returns NULL on a valid untied hash - it will always succeed, or
croak if memory allocation fails.

This is a regression introduced in 2016 as part of commit a50a854f07ccc665:
    Fix %DB::sub for subs with non-Latin1 characters

That commit refactored the code from hv_store to hv_store_ent - a sensible
choice, as the latter API is more efficient - but introduced two small
errors not present in the previous code - inverting sense the truth test on
the return from hv_store_ent, and changing a SvREFCNT_dec to a SvREFCNT_inc.
Hence it would "leak" one reference count.

Experimentation suggests that this code doesn't cause obvious problems if it
is actually wrong. The code is adding another key that points to a value
already in the hash, and (intending) to have the correct reference count
manipulations to reflect this. Because all entries in the hash persist until
interpreter exit, and nothing deletes from the hash, in general "leaking" a
reference count won't make a difference as the SV already persists for the
lifetime of the interpreter. Likewise, deliberately introducing a different
bug into this code and *not* increasing the reference count also triggers no
warnings or errors, because the existing entry in the hash keeps the (new,
aliased) value "alive" for the lifetime of the hash.

Hence why this hasn't been noticed.

### DIFF
--- a/ListUtil.xs
+++ b/ListUtil.xs
@@ -2029,12 +2029,13 @@ PPCODE:
         }
 
         if (old_data && HeVAL(old_data)) {
+            SV* old_val = HeVAL(old_data);
             SV* new_full_name = sv_2mortal(newSVpvn_flags(HvNAME(stash), HvNAMELEN_get(stash), HvNAMEUTF8(stash) ? SVf_UTF8 : 0));
             sv_catpvn(new_full_name, "::", 2);
             sv_catpvn_flags(new_full_name, nameptr, s - nameptr, utf8flag ? SV_CATUTF8 : SV_CATBYTES);
-            SvREFCNT_inc(HeVAL(old_data));
-            if (hv_store_ent(DBsub, new_full_name, HeVAL(old_data), 0) != NULL)
-                SvREFCNT_inc(HeVAL(old_data));
+            SvREFCNT_inc(old_val);
+            if (!hv_store_ent(DBsub, new_full_name, old_val, 0))
+                SvREFCNT_dec(old_val);
         }
     }
 


nwc10:
@Leont commit a50a854f07ccc665 was yours. I think I have figured things out correctly - could you check?

(Seems github doesn't give me permission to suggest you as a reviewer.)

haarg:
Is there a reasonable way this could be tested?

Leont (approved):

Leont:
The double increment is definitely a bug

nwc10:
> Is there a reasonable way this could be tested?

Good question. I'd not really considered this. So I went for a dig...

The variable - the hash - isn't even documented. intrpvar.h says

```
=for apidoc mn|GV *|PL_DBsub
When Perl is run in debugging mode, with the B<-d> switch, this GV contains
the SV which holds the name of the sub being debugged.  This is the C
variable which corresponds to Perl's $DB::sub variable.  See
C<L</PL_DBsingle>>.
```

but looking at the way the GV is initialised, I can see that it's exposed as `%DB::sub`. But it only exists (well, is populated) under the debugger.

So I think to test it you'd need to

1) shell out to the same perl that you're building against, but with `-d` (and `blib` set up correctly)
2) either with the environment set up to force the debugger to run particular commands, or (maybe more easily) with a custom debugger module somewhere in `t/lib` that does what is needed
3) Running a script that does some sort of call to `set_subname`, and then inspects the reference counts in `%DB::sub`
4) And knows what the "right" answer is (for that version of perl, if it changes between versions)
5) But does fail with the wrong code in our XS

which feels convoluted and extremely fragile.

haarg:
Here's a oneliner that can demonstrate the difference this patch makes:
```
$ perl -E'use Sub::Util (); $^P |= (0x10); eval q{ sub namedsub { 1 }; 1 } or die $@; Sub::Util::set_subname("newnamedsub", \&namedsub ); say Internals::SvREFCNT($DB::sub{"main::newnamedsub"});'
3
$ perl -Mblib -E'use Sub::Util (); $^P |= (0x10); eval q{ sub namedsub { 1 }; 1 } or die $@; Sub::Util::set_subname("newnamedsub", \&namedsub ); say Internals::SvREFCNT($DB::sub{"main::newnamedsub"});'
2
```
It still may not be worth the effort to add a test.

nwc10:
Oooh - thanks for figuring out a way to eliminate some of the things I thought would be necessary, and hence get it down to a oneliner.

I'm also still not really confident that it's worth the effort to add a test, ensure that it skips or passes on all perl versions, and tests anything broader than "we don't repeat exactly this mistake again".

PR #116, opened 2021-09-06
The minimum Perl version is v5.6.0, which supports weak references

nwc10:
Hence remove all the conditional logic and documentation for supporting
earlier versions of Perl which did not have weak references.


leonerd:
Thanks. Merged in fc41f80

PR #120, opened 2022-03-06
Skip the tainted tests if your perl was built without taint support

neilb:
Hi,

I'm working on changes to the Configure script, to make it easy to build a perl that doesn't have taint support.

This PR changes `t/taint.t` to skip all tests if being installed on such a perl.

Cheers,
Neil

### DIFF
--- a/t/tainted.t
+++ b/t/tainted.t
@@ -3,9 +3,17 @@
 use strict;
 use warnings;
 
-use Test::More tests => 5;
-
+use Config;
+use Test::More;
 use Scalar::Util qw(tainted);
+
+if (exists($Config{taint_support}) && not $Config{taint_support}) {
+    plan skip_all => "your perl was built without taint support";
+}
+else {
+    plan tests => 5;
+}
+
 
 ok( !tainted(1), 'constant number');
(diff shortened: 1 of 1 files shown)


neilb:
I've submitted a [pull request on Perl](https://github.com/Perl/perl5/pull/19541) which adds a Configure question for taint support.

Given Scalar-List-Utils is dual-life, but CPAN upstream, I included these changes in my PR on perl, but marked it "do not merge", as the right path forward is for a release of Scalar-List-Utils, which is then merged with blead.

Can you check the one changed file please, and if you're happy with the change, do a release to CPAN, and then merge into blead as well?

leonerd (approved):
LGTM

leonerd:
Now merged. I want to make some more documentation edits to point users at the new `builtin::` core functions before I do another release. Hopefully today/tomorrow.

leonerd:
Now released on CPAN as 1.62. Once that's bounced around the mirrors that should be fine to pull into bleadperl.

https://metacpan.org/release/PEVANS/Scalar-List-Utils-1.62

PR #124, opened 2023-03-26
Skip taint tests when -DSILENT_NO_TAINT_SUPPORT is enabled.

DrHyde:
If perl is built with -DSILENT_NO_TAINT_SUPPORT checking `$Config{taint_support}` isn't enough, this adds a check that looks for -DSILENT_NO_TAINT_SUPPORT in `$Config{ccflags}`.


DrHyde:
closing, because of discussions on the related perl5-blead PR

PR #130, opened 2024-03-29
Don't loop forever in uniqnum.t

craigberry:
If infinities are not working, make sure we fail after a few tries rather than looping infinitely to find infinity. 10 iterations is arbitrary -- 1 is likely enough in most cases.

### DIFF
--- a/t/uniqnum.t
+++ b/t/uniqnum.t
@@ -190,7 +190,10 @@ is_deeply( [ uniqnum @in],
 # Hard to know for sure what an Inf is going to be. Lets make one
 my $Inf = 0 + 1E1000;
 my $NaN;
-$Inf **= 1000 while ( $NaN = $Inf - $Inf ) == $NaN;
+for (1..10) {
+    $Inf **= 1000;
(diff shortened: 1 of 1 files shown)


jkeenan:
The code in this pull request has been shipping with the Perl core distribution for several months now.
```
commit a49eddd88634f28613f99206b3279954eb01d8da
Author:     Craig A. Berry <craigberry@mac.com>
AuthorDate: Fri Mar 29 13:32:19 2024 -0500
Commit:     Craig A. Berry <craigberry@mac.com>
CommitDate: Fri Mar 29 17:10:19 2024 -0500

    Don't loop forever in uniqnum.t
```
However, that means that in the core distribution we have to designate cpan/Scalar-List-Utils/t/uniqnum.t as a `CUSTOMIZED` file, which poses some administrative difficulty.  Could we get the commit merged into this distro's `master` branch and get a new CPAN release (which we would then sink into Perl `blead`)?  Thanks.

leonerd:
Oh oops - I had entirely missed this one. I'll take a look now.

PR #131, opened 2024-04-27
Adjust testsuite for perl-versions

atoomic:
Remove macOS and windows pipeline as they are
discouraged by GitHub as they consume more ressource and brings a low value.

Use the `perl-actions/perl-versions@v1` pipeline to use a dynamic list of current Perl versions.


leonerd (approved):
LGTM

PR #132, opened 2024-06-12
Use builtin:: functions where available

leonerd:
On Perl 5.40 and above, use the various `builtin::` functions to provide those in `Scalar::Util`, avoiding the need for our own custom xsubs in that case

### DIFF
--- a/ListUtil.xs
+++ b/ListUtil.xs
@@ -1717,6 +1717,7 @@ CODE:
     ST(0) = boolSV((SvPOK(sv) || SvPOKp(sv)) && (SvNIOK(sv) || SvNIOKp(sv)));
     XSRETURN(1);
 
+#if !PERL_VERSION_GE(5, 40, 0)
 SV *
 blessed(sv)
     SV *sv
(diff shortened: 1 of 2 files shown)


karenetheridge:
Some of these functions were available in 5.38's builtin, so should we start with that version?

leonerd:
> Some of these functions were available in 5.38's builtin, so should we start with that version?

They're present in 5.38 (or even 5.36) but in those versions they print experimental warnings. It isn't until 5.40 when they become stable.

karenetheridge:
Did they change between 5.38 and 5.40? If they didn't, we could retroactively consider them essentially stable in 5.38, in the same way that the `stable` declaration does in `experimental.pm`. (Therefore, for 5.38, the Scalar::Util shim would need to silence the experimental warning.)

haarg:
The functions would need a wrapper sub to be able to silence the warning.

leonerd:
Yeah, there's only one warning category, not one per function, so it's not possible to silence just the now-stable functions. I don't see it as a huge problem though - the XS versions will have to live here until we decide to drop support for anything older than 5.40 and I don't imagine that happening for a decade or two at least ;) Whether the horizon is 5.38 or 5.40 is no huge problem between those.

leonerd:
There doesn't seem to be any overriding reason or argument not to do this, so if no objections by Aug 1st I shall merge it.

PR #133, opened 2024-08-12
only treat ' as special in versions of perl where it's special

tonycoz:


leonerd:
perl 5.6 CI box appears unhappy in a way I think is unrelated to this change. Mostly GLIBC upsets. I shall ignore it and merge anyway

PR #135, opened 2024-10-16
add Scalar::List::Utils module to match dist name

haarg:
Various tools expect to be able to translate distribution names to module names. That isn't always true, but it's helpful when it can be.

Add a Scalar::List::Utils module to match the dist name, including links to the other included modules.


leonerd:
Doesn't it want to be "Utils", plural?

haarg:
Good catch. Fixed.

leonerd (approved):
LGTM

PR #137, opened 2025-01-10
Fix uniqint SIGSEGV for objects with no int overload

zhmylove:
In case uniqint takes several objects, it always tries to call their int overload using amagic_call.  This patch fixes the case when amagic_call returned NULL resulting into Segmentation Violation. Several tests are also added to ensure proper objects handling.

**Problem:**
The code below expected to work but crashes with SIGSEGV.
```perl
use List::Util qw( uniqint );
uniqint bless {}, __PACKAGE__; # given __PACKAGE__ has no 'int' overload unlike Math::BigInt
```

**Desired behaviour:**
Returns a list with single reference to an object.

### DIFF
--- a/ListUtil.xs
+++ b/ListUtil.xs
@@ -1369,8 +1369,9 @@ CODE:
             /* coerce to integer */
 #if PERL_VERSION >= 8
             /* int_amg only appeared in perl 5.8.0 */
-            if(SvAMAGIC(arg) && (arg = AMG_CALLun(arg, int)))
-                ; /* nothing to do */
+            SV *tmpsv;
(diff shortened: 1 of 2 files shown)


zhmylove:
@leonerd may I ask you to take a look on the PR as it fixes a segfault in `uniqint()` please

zhmylove:
Anyone?

zhmylove:
Perl5.6 check fails due to CI issue

sisyphus:
Consider posting to the p5p mailing list about this lack of response.
It's an issue that should be addressed, and someone on that mailing list (who is not even currently aware of this issue) might be able to initiate some action.

zhmylove:
I decided to use croak instead of implicit integer cast.
Opened #138

PR #138, opened 2025-01-26
Fix uniqint SIGSEGV for objects with no int overload

zhmylove:
In case uniqint takes several objects, it always tries to call their int overload using amagic_call.  This patch fixes the case when amagic_call returned NULL resulting into Segmentation Violation. Several tests are also added to ensure proper objects handling.

**Problem:**
The code below expected to work but crashes with SIGSEGV.
```perl
use List::Util qw( uniqint );
{ package Foo; use overload "+" => sub {1} }
uniqint bless {}, "Foo"; # Foo has no 'int' overload unlike Math::BigInt
```

```
| SvAMAGIC | AMG_CALLun | Current result | Proposed result |
+----------+------------+----------------+-----------------+
| false    | false      | else branch    | else branch     |
| false    | true       | else branch    | else branch     |
| true     | false      | SIGSEGV        | croak like int()|
| true     | true       | else bypassed  | else bypassed   |
```

### DIFF
--- a/ListUtil.xs
+++ b/ListUtil.xs
@@ -1369,8 +1369,10 @@ CODE:
             /* coerce to integer */
 #if PERL_VERSION >= 8
             /* int_amg only appeared in perl 5.8.0 */
-            if(SvAMAGIC(arg) && (arg = AMG_CALLun(arg, int)))
-                ; /* nothing to do */
+            if(SvAMAGIC(arg)) {
+                if(!(arg = AMG_CALLun(arg, int)))
+                    croak("No \"int\" method found in overloaded package");
+            }
             else
 #endif
             if(!SvOK(arg) || SvNOK(arg) || SvPOK(arg))
--- a/t/uniq.t
+++ b/t/uniq.t
@@ -3,7 +3,7 @@
 use strict;
 use warnings;
 use Config; # to determine ivsize
-use Test::More tests => 31;
+use Test::More tests => 35;
 use List::Util qw( uniqstr uniqint uniq );
 
 use Tie::Array;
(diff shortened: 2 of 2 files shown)


ListUtil.xs
```diff
@@ -1369,8 +1369,10 @@ CODE:
             /* coerce to integer */
 #if PERL_VERSION >= 8
             /* int_amg only appeared in perl 5.8.0 */
-            if(SvAMAGIC(arg) && (arg = AMG_CALLun(arg, int)))
-                ; /* nothing to do */
+            if(SvAMAGIC(arg)) {
+                if(!(arg = AMG_CALLun(arg, int)))
```
leonerd, line 1373: Hmm. This adds a new exception condition where there didn't used to be one. I wonder if that change in behaviour would break too much.

I'd prefer to preserve the existing condition, but without trashing the `arg` value. Instead, perhaps have:

```c
SV *mgret;
if(SvAMAGIC(arg) && (mgret = AMG_CALLun(arg, int)))
    arg = mgret;
else ...
```

that way, it keeps the existing condition, but does not trash the value in `arg` when there is some magic but no int_amg.

zhmylove: Thanks!
Yes, I actually suggested this particular change in #137 

But after that I thought a little bit more about that and decided to write this with `croak` just like `int()` does in this situation. 

This change will definitely not break any existing code as it croaks only instead of Segmentation Violation error which causes exit even inside `eval`

leonerd: Ahyes. In that case, does the message match what would happen normally?

leonerd:
The Perl 5.6 smoker looks unhappy, but looks like a machine setup issue, unrelated to this change.

ListUtil.xs
haarg: The message does not match.

But looking at the code for `amagic_call`, it isn't obvious why the previous code isn't already working correctly. It seems like it should already be producing an appropriate croak if the overload doesn't exist.

zhmylove: @leonerd > what would happen normally

In 5.42.0, just like in the older versions, it still results into a crash with segmentation violation.

Typical error of missing `"int"` overload looks like:
```
Operation "+": no method found,
	left argument has no overloaded magic,
	right argument in overloaded package OverloadedPackageWithoutInt at -e line 1.
```

To me, this does not look meaningful in case of `uniqint()`.

Maybe we should merge this PR to avoid segmentation faults in Perl core modules?
In my solution the message tells `No "int" method found in overloaded package` and highlights the line with `uniqint()` call.

PR #139, opened 2025-03-29
always allow isvstring to be used

haarg:
On perl 5.6, the isvstring function would not allow itself to be called or exported. This behavior is annoying to implement, and forces loading Exporter::Heavy in all cases.

Instead, isvstring can just always return false on perl 5.6. This is an accurate behavior, as nothing on perl 5.6 is a vstring. And the odds of anyone caring about this specific behavior on a prehistoric version of perl is almost certainly 0.


haarg:
I've tested locally using perl 5.6 and it passes its tests.

leonerd:
Then I'll take it :)

PR #141, opened 2025-07-14
re-instate ' in names

tonycoz:


t/exotic_names.t
```diff
@@ -45,7 +45,7 @@ sub caller3_ok {
         ),
     );
 
-    $expected =~ s/'/::/g if $] < 5.041_003;
+    $expected =~ s/'/::/g if $] < 5.037009 || $] >= 5.041_004;
```
karenetheridge, line 48: If the code will work for all versions, then we should remove the conditional here, so as to test on all versions, rather than add more gating?

tonycoz: It looks like this condition could be removed safely.

The conditionals further down (for `@ordinal` and `$legal_ident_char`) are needed to prevent warning from 5.38/5.40:
```
# 5.40
t/dualvar.t ........... ok
t/exotic_names.t ...... 1/1560 Old package separator "'" deprecated at (eval 285) line 1.
t/exotic_names.t ...... ok
```

I don't know if we care here about 5.41.x dev releases now 5.42 has been released, but a range of 5.41.x will fail without the conditionals.

leonerd: I wouldn't worry if new releases break against previous development cycles in the middle, for the sake of keeping conditions simpler. We may already have other things that break mid-cycle developments anyway.

karenetheridge: I misunderstood what the conditional would actually be doing.

If removing it would cause test breakage on development releases, I would prefer to leave it in.  Dev releases are occasionally still run for bisection purposes, so it would be nicer if tests didn't fail there.

karenetheridge:
Also, do we need to check if `feature "apostrophe_as_package_separator"` has been disabled?

tonycoz:
> Also, do we need to check if `feature "apostrophe_as_package_separator"` has been disabled?

I don't think it matters.

`'` in names is always valid in symbolic references and treated as `::`, whether the feature is enabled or not[1], and what set_subname() is doing is setting the symbolic name of the CV, and when `'` is allowed in names in source, it's always converted to `::`.

[1] see https://github.com/Perl/perl5/issues/22644#issuecomment-2398064628 onwards

PR #142, opened 2025-07-29
re-instate ' in names, from 5.41.6 and later

karenetheridge:
- ' was deprecated in 5.37.9, so we stopped testing for it, but code still supported it
- ' was removed entirely in 5.41.3, so we stopped supporting it in code
- ' was reinstated entirely in 5.41.6 (but behind a default-enabled feature), so we can support it in code and test for it again

This replaces #141 and https://github.com/Perl/perl5/pull/23500 -- after this is released as version 1.71, blead can then sync to it.

### DIFF
--- a/ListUtil.xs
+++ b/ListUtil.xs
@@ -1951,7 +1951,7 @@ PREINIT:
     STRLEN namelen;
     const char* nameptr = SvPV(name, namelen);
     int utf8flag = SvUTF8(name);
-#if PERL_VERSION_LT(5, 41, 3)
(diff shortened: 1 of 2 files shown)


karenetheridge:
I think this is not quite right still as it ignores the existence of the `apostrophe_as_package_separator` feature -- how can I check for its state in XS? when the feature is off we should not be enabling the special `'` handling code.

karenetheridge:
The testsuite fails on 5.6, but this looks like a workflow problem?

leonerd:
Yeah; the 5.6 VM looks unhappy, not at this change. I expect this will be fine.

leonerd (approved):
LGTM.

PR #143, opened 2025-08-17
uniqnum: Update and remove undefined C behavior

khwilliamson:
1.  Configure for all but ancient perls tells us if a UV can always losslelssly fit into an NV.   Use that if available.
2.  Configure for all but ancient perls tells us the mantissa width.  Use that if available instead of assuming it is 53.
3.  The code for `uniqnum` assumes this is a 2's complement machine.  Change to not assume that.
4. There were several undefined behaviors marked by running ASan on this code; all stemmed from using signed values, where unsigned would work.   There was an issue with abs(IV_MIN) not being representable in an IV on  a 2's complement machine; and left shifting a negative value is undefined.   
5.  I found myself drawing a picture to try to understand the behavior of  item 4.  And I think the picture is clearer than the previous comments. so I     changed to use it.

### DIFF
--- a/ListUtil.xs
+++ b/ListUtil.xs
@@ -1458,6 +1458,9 @@ CODE:
             SvNV(arg); /* SvIV() sets SVf_IOK even on floats on 5.6 */
 #endif
         }
+#ifndef NV_PRESERVES_UV_BITS
+#  define NV_PRESERVES_UV_BITS 53
+#endif
 #if PERL_VERSION < 6
 #  if NVSIZE > IVSIZE               /* $Config{nvsize} > $Config{ivsize} */
 #    define NV_PRESERVES_UV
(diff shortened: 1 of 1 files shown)


mauke:
> The code for `uniqnum` assumes this is a 2's complement machine. Change to not assume that.

NB: Starting with C23, C requires two's complement.

sisyphus:
This PR builds and tests fine for me on MS Windows, on a wide variety of perls across versions 5.16.3 to 5.43.1 - including 32-bit, 64-bit, -Duselongdouble, -Dusequadmath.
It was only 51 builds, mainly focusing on builds from over the last 3 years - so there are plenty of holes (some of them quite large) in the coverage.

One thing I noticed was that t/uniqnum.t produced the following warning (but only in the 5.42.0 and 5.43.1 builds) :
```
Possible precedence problem between ! and numeric ne (!=) at t/uniqnum.t line 215.
```
I don't think that's even related to the changes made by this PR, but line 215 is:
```
$nanish != 0 && !$nanish != $NaN;
```
I was able to silence the warning by changing that line to
```
$nanish != 0 && (!$nanish) != $NaN;
```
I think that retains the original (and intended) behaviour.

leonerd:
Gosh; that original code is kinda crazy - it really needs a first-line "hey here's what we're doing" to explain what is going on... but anyhow...

Yes I think this all looks fine, but the 5.6 check fails on it. The logs have now been deleted so we can't see why. Is there a way to rerun this?

